PR #58, opened 2019-01-20
use Adj(x\inv) rather than Adj(x)\inv

artivis:


joansola:
Ah ok i see. Good to merge

joansola:
I suppose in my comment above that the ompiler is aliminating any temporary storage, and so that what is actually done is
```c++
*J_mc_ma = m_SO3.inverse().rotation().
```

PR #60, opened 2019-01-30
Deprecate lift & retract

artivis:
Deprecate 
- `lift` in favor of `log`
- `retract` in favor of `exp`


joansola:
Yep I can see that.

For me this MR is OK. Are you waiting for my review? In case you are, I approved yesterday.

PR #61, opened 2019-01-30
Fix random

artivis:
Fix `random` which was spanning range `[-1, 1]` for all `Tangent` coefficients.
It now spans range `[-pi, pi]` where necessary.

The current implementation requires a specialization for each intermediate `Tangent` derived types.

### DIFF
--- a/include/manif/impl/se2/SE2Tangent_base.h
+++ b/include/manif/impl/se2/SE2Tangent_base.h
@@ -335,6 +335,17 @@ struct WEvaluator<SE2TangentBase<Derived>>
   }
 };
 
+template <typename Derived>
+struct RandomEvaluatorImpl<SE2TangentBase<Derived>>
+{
+  template <typename EigenDerived>
+  static void run(Eigen::MatrixBase<EigenDerived>& m)
+  {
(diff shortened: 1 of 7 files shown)


include/manif/impl/se2/SE2Tangent_base.h
```diff
@@ -335,6 +335,17 @@ struct WEvaluator<SE2TangentBase<Derived>>
...
 };
 
+template <typename Derived>
+struct RandomEvaluatorImpl<SE2TangentBase<Derived>>
+{
+  template <typename EigenDerived>
+  static void run(Eigen::MatrixBase<EigenDerived>& m)
+  {
+    // in [-1,1] /   in [-PI,PI]
+    m.setRandom().coeffRef(2) *= M_PI;
```
joansola: Is it necessary to have this one liner? Readability-wise, I'd prefer a two-liner.

artivis: No it isn't, I'll update it.

artivis: Done.

PR #63, opened 2019-02-03
Optimize code for SE3 ljac and rjac

joansola:
Found a shortcut in Barfoot's formula (see comment in code). 

Then also made some temporaries to save some extra CPU time.

### DIFF
--- a/include/manif/impl/se3/SE3Tangent_base.h
+++ b/include/manif/impl/se3/SE3Tangent_base.h
@@ -130,8 +130,7 @@ public:
 
 protected:
 
-  template <typename _EigenDerived>
-  void fillQ(Eigen::MatrixBase<_EigenDerived>& Q) const;
+  void fillQ(Eigen::Ref<Eigen::Matrix<Scalar, 3, 3>> Q) const;
 };
 
 template <typename _Derived>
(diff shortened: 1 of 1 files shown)


include/manif/impl/se3/SE3Tangent_base.h
```diff
@@ -198,12 +198,18 @@ SE3TangentBase<_Derived>::rjac() const
   }
 
   /// @note Barfoot14tro Eq. 102
+  const Eigen::Matrix<Scalar, 3, 3> VW  = V * W;
+  const Eigen::Matrix<Scalar, 3, 3> WV  = VW.transpose();       // Note on this change wrt. Barfoot: it happens that V*W = (W*V).transpose() !!!
+  const Eigen::Matrix<Scalar, 3, 3> WVW = WV * W;
+  const Eigen::Matrix<Scalar, 3, 3> VWW = VW * W;
+  const Eigen::Matrix<Scalar, 3, 3> WWV = - VWW.transpose();    // Note on this change wrt. Barfoot: it happens that V*W*W = -(W*W*V).transpose() !!!
```
artivis: Do we really need this temporary ? It is used only once.
Same remark in Jl.

joansola: Which one, WWV? No, it could go straight into the formula. Change it if you wish, but keep the comment otherwise we'll  lose track.

joansola: OK I do it

joansola: done and pushed

PR #64, opened 2019-02-04
Add example of SAM in SE2

joansola:
It's all in the title


artivis:
@joansola Force-pushed after rebase on `devel`.
Some small fixes & cosmetics.

PR #65, opened 2019-02-06
Fix/instream

artivis:
Fix Tangent input stream operator. It simply forwards the input stream to Eigen.


test/se2/gtest_se2_tangent.cpp
```diff
@@ -103,6 +103,30 @@ TEST(TEST_SE2, TEST_SE2_TANGENT_RETRACT_JAC)
 //  EXPECT_DOUBLE_EQ(1, J_ret(0));
 }
 
+TEST(TEST_SE2, TEST_SE2_TANGENT_INSTREAM)
```
artivis: @joansola FYI

PR #69, opened 2019-03-09
fix #68 - quaternion randomization

### LINKED ISSUE #68: Improve Randomization
Manif uses random tangent space element and expMap to generate a random Lie group element:
[setRandom](https://github.com/artivis/manif/blob/devel/include/manif/impl/lie_group_base.h#L370) `coeffs_nonconst() = Tangent::Random().exp().coeffs()`

This is not a good idea. Because uniform distribution of tangent space may not lead to uniform distribution of Lie group.

I used the following code to test the Random() function in SO(3) by generating 10000 random elements:

```
// generate_so3_random_elements.cpp
#include "manif/SO3.h"

#include <vector>
#include <iostream>
#include <fstream>
#include <Eigen/Dense>

using namespace manif;
using namespace std;

int main(int /*argc*/, char ** /*argv*/)
{
    ofstream file;
    file.open("SO3Randomization.txt");
    for (int i = 0; i < 10000; ++i)
    {
        SO3d X = SO3d::Random();
        file << X.coeffs().transpose() << endl;
    }
    file.close();

    return 0;
}
```

and plotted the axis distribution (on a unit sphere) and angle distribution (with a histogram) in Matlab:

```
%% plot_so3_distribution.m
clear; clc; close all;
qua = importdata('SO3Randomization.txt');
axang_arr = quat2axang(qua);
figure(1)
plot3(axang_arr(:,1), axang_arr(:,2), axang_arr(:,3), '.');
figure(2)
hist(axang_arr(:,4),100); 
```

The Random() function cannot generate uniformly distributed SO3 elements.

artivis:
Fix issue #68 by using `Eigen::Quaternion<Scalar>::UnitRandom()` in the randomization of SO3 and SE3.

### DIFF
--- a/include/manif/impl/se3/SE3_base.h
+++ b/include/manif/impl/se3/SE3_base.h
@@ -339,6 +339,21 @@ SE3Base<_Derived>::z() const
   return coeffs().z();
 }
 
+namespace internal {
+
+template <typename Derived>
+struct RandomEvaluatorImpl<SE3Base<Derived>>
+{
+  template <typename T>
+  static void run(T& m)
+  {
+    using Translation = typename SE3Base<Derived>::Translation;
+    using Quaternion  = typename SE3Base<Derived>::QuaternionDataType;
+    m = Derived(Translation::Random(), Quaternion::UnitRandom());
+  }
+};
+
+} /* namespace internal */
 } /* namespace manif */
 
 #endif /* _MANIF_MANIF_SE3_BASE_H_ */
--- a/include/manif/impl/so3/SO3_base.h
+++ b/include/manif/impl/so3/SO3_base.h
@@ -337,6 +337,20 @@ void SO3Base<_Derived>::normalize()
   coeffs().normalize();
 }
 
+namespace internal {
+
+template <typename Derived>
+struct RandomEvaluatorImpl<SO3Base<Derived>>
+{
+  template <typename T>
+  static void run(T& m)
+  {
+    using Quaternion = typename SO3Base<Derived>::QuaternionDataType;
+    m = Derived(Quaternion::UnitRandom());
+  }
+};
+
+} /* namespace internal */
 } /* namespace manif */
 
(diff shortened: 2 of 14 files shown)


include/manif/impl/se3/SE3_base.h
```diff
@@ -339,6 +339,21 @@ SE3Base<_Derived>::z() const
...
+
+template <typename Derived>
+struct RandomEvaluatorImpl<SE3Base<Derived>>
+{
+  template <typename T>
+  static void run(T& m)
+  {
+    using Translation = typename SE3Base<Derived>::Translation;
+    using Quaternion  = typename SE3Base<Derived>::QuaternionDataType;
+    m = Derived(Translation::Random(), Quaternion::UnitRandom());
```
artivis: @joansola See here.

include/manif/impl/so3/SO3_base.h
```diff
@@ -337,6 +337,20 @@ void SO3Base<_Derived>::normalize()
...
+namespace internal {
+
+template <typename Derived>
+struct RandomEvaluatorImpl<SO3Base<Derived>>
+{
+  template <typename T>
+  static void run(T& m)
+  {
+    using Quaternion = typename SO3Base<Derived>::QuaternionDataType;
+    m = Derived(Quaternion::UnitRandom());
```
artivis: @joansola See here.

artivis:
@joansola I couldn't run the matlab script from #68, can you please give it a shot with this fix?

joansola:
which matlab script???

artivis:
The few plotting lines provided by qqfly [here](https://github.com/artivis/manif/issues/68#issue-411295677)

joansola:
But these plot some results that have to be generated from a c++ implementation using manif. Did you generate a `.txt` file with results so that I can plot them in Matlab?

joansola:
image
image
All OK

PR #70, opened 2019-03-13
Fix/doc

artivis:
This PR mostly fixes some comments formatting which was interfering with the generation of the documentation.  
It also adds a link the online API doc in the readme.


artivis:
@joansola This PR does not change any code - but a few debugging-related cout in a separate function.  
It only fixes the formatting of comments that was interfering with the generation of the documentation.

PR #71, opened 2019-03-21
Fix Eigen 3.3.6 bug

qqfly:
There is a bug in Eigen 3.3.6: [Bug 1643 - Compilation failure](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1643).

This bug occurs when building `examples/se2_localization.cpp` and `examples/se3_localization.cpp`

### DIFF
--- a/cmake/modules/FindEigen3.cmake
+++ b/cmake/modules/FindEigen3.cmake
@@ -58,7 +58,7 @@ macro(_eigen3_check_version)
 
   if(NOT EIGEN3_VERSION_OK)
     if(${EIGEN3_VERSION} VERSION_EQUAL "3.3.6")
-      message(FATAL_ERROR "Eigen3 version 3.3.6 found in /home/rvbust/Rvbust/Install/Eigen/include/eigen3/,"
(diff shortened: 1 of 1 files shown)


artivis:
Hi,
Thanks for the PR, I'll have a look at it over the week-end :+1:

artivis:
@qqfly You have edited directly the `FindEigen3.cmake` file which comes directly from upstream Eigen. Rather than this I would prefer to add a simple warning in `manif` base `CMakeLists.txt`.
Something like :

```cmake
find_package(Eigen3 REQUIRED)

if(${EIGEN3_VERSION} VERSION_EQUAL "3.3.6")
  # We only warn about this problem and delegate to Eigen whether to crash the compilation or not.
  message(WARNING "Eigen3 version ${EIGEN3_VERSION} found in ${EIGEN3_INCLUDE_DIR},"
                  "but this version has a [bug](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1643)")
endif(${EIGEN3_VERSION} VERSION_EQUAL "3.3.6")
```

Do you mind updating your PR ?

qqfly:
The source branch has been deleted, so I'll close this PR, and issue a new PR.

PR #72, opened 2019-03-25
Fix const SO3 normalize()

qqfly:
1. const coeffs() cannot be normalized [code](https://github.com/artivis/manif/blob/devel/include/manif/impl/so3/SO3_base.h#L335)

2. SO3 constructor should do the normalization (?)  **OR** make normalize() a public function.

3. #include <Eigen/Geometry> should be added to support Eigen::Quaternion.

### DIFF
--- a/include/manif/impl/so3/SO3_base.h
+++ b/include/manif/impl/so3/SO3_base.h
@@ -4,6 +4,7 @@
 #include "manif/impl/so3/SO3_properties.h"
 #include "manif/impl/lie_group_base.h"
 #include "manif/impl/utils.h"
+#include <Eigen/Geometry>
 
 namespace manif {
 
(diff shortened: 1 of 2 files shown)


include/manif/impl/so3/SO3_base.h
```diff
@@ -332,7 +333,7 @@ SO3Base<_Derived>::quat() const
 template <typename _Derived>
 void SO3Base<_Derived>::normalize()
 {
-  coeffs().normalize();
+  coeffs_nonconst().normalize();
```
artivis: @qqfly good catch :+1:

artivis:
I am not sure yet whether we should allow constructing a `SO3` object from un-normalized quaternion or not. We could, and normalize it under the hood, but that's hiding it from the user. Or we could not allow it, test for normalization and throw an error in case it is not. So far I personally favor the second option.
@joansola Any take on this ??

joansola:
I seem to favor also the second. However, this test should only be made once at construction time, otherwise it's performance bloat.

artivis:
@qqfly Sorry this took so long. I am going to close this PR since we decided that arguments provided to a constructor should be already normalized. In case they are not, it throws an exception (See #93).
Thanks again for your interest in this work.

PR #73, opened 2019-03-26
Fix Eigen 3.3.6 bug

qqfly:
There is a bug in Eigen 3.3.6: [Bug 1643 - Compilation failure](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1643).

This bug occurs when building `examples/se2_localization.cpp` and `examples/se3_localization.cpp`


artivis:
Thanks for the update @qqfly :+1:

PR #74, opened 2019-03-26
Fix/eigen version

artivis:
`Eigen::Quaterniond::UnitRandom()` is not available to Eigen 3.3-beta1 which seems to be the default Eigen version of Ubuntu 16.04.
- `UnitRandom` function implementation is copied directly in `manif` to improve compatibility
- Travis CI uses now Eigen 3.3-beta1


artivis:
It is a copy-paste ^^

PR #75, opened 2019-04-09
Documentation fix & improvements

artivis:
Adding `.codedocs` config file fixes the online documentation configuration at [codedocs.xyz/artivis/manif](https://codedocs.xyz/artivis/manif/).

### DIFF
--- a/.codedocs
+++ b/.codedocs
@@ -1,0 +1,1 @@
+DOXYFILE=.doxygen.txt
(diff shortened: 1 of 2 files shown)


artivis:
@joansola Can you please have a look at this PR which adds a couple forewords (copy/paste from joss paper) about `manif`.  
Also have a look at the updated [ceres-related wiki](https://github.com/artivis/manif/wiki/On-the-use-with-Ceres), I have troubles on Eq. `(3)`.

PR #77, opened 2019-04-25
Add constructor from eigen::isometry in SE3

joansola:
Add one new constructor in `SE3`, and the corresponding unit test.

This constructor uses `Eigen::Affine3d` as the input parameter, which is pretty equivalent to an SE3 object: one rotation and one translation.

Note on `Eigen::Affine3d`: `Affine3d` is an alias of `Transform3d`, just that the linear part is marked as a rotation matrix in order to speed up the `inverse()` procedure: we will transpose the linear part assuming it's a rotation matrix, instead of inverting it assuming it is just a matrix.

### DIFF
--- a/include/manif/impl/se3/SE3_base.h
+++ b/include/manif/impl/se3/SE3_base.h
@@ -5,6 +5,8 @@
 #include "manif/impl/lie_group_base.h"
 #include "manif/impl/so3/SO3_map.h"
 
+#include <Eigen/Geometry>
+
 namespace manif {
 
 //
@@ -34,6 +36,7 @@ public:
   using Rotation       = typename internal::traits<_Derived>::Rotation;
   using Translation    = typename internal::traits<_Derived>::Translation;
   using Transformation = typename internal::traits<_Derived>::Transformation;
+  using Isometry       = Eigen::Transform<Scalar, 3, Eigen::Isometry>;
 
   using QuaternionDataType = Eigen::Quaternion<Scalar>;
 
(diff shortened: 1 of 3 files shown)


artivis:
LGTM.

The same manner we can construct a SE3 object from a 3D Isometry, it may be interesting to be able to retrieve a 3D Isometry from a SE3 object, e.g.
```cpp
Eigen::Isometry3d = se3.isometry();
```
Notice that there is already a similar function in SE3Base, [SE3Base::transform()](https://github.com/artivis/manif/blob/1e29fe00bca6f6d7cdaeb223e02be8421abda9e9/include/manif/impl/se3/SE3_base.h#L107).

joansola:
I think transform() should rather return an Isometry in this case.

The diff between transform and isometry is just a marker to speed up things. Coming from SE3, this mark should be set as to indicate that the matrix part is a rotation matrix

joansola:
But OK, we can add a isometry() method and be explicit in what we do.

include/manif/impl/se3/SE3_base.h
```diff
@@ -181,9 +181,8 @@ template <typename _Derived>
 typename SE3Base<_Derived>::Isometry
 SE3Base<_Derived>::isometry() const
 {
-  Isometry T = Isometry::Identity();
-  T.matrix().template topLeftCorner<3,3>()  = rotation();
-  T.matrix().template topRightCorner<3,1>() = translation();
+  Isometry T;
+  T.matrix() = transform();
   return T;
```
artivis: Maybe this can even be simplified further :
```cpp
return Isometry(transform());
```

PR #80, opened 2019-05-18
Fix #78 about Jacobians in se[2,3]_sam.cpp

### LINKED ISSUE #78: Question about Jacobian in example se2_sam/se3_sam.cpp
Hi, thanks for your work!
In your example `se2_sam.cpp`(or `se3_sam.cpp`), for the prior measurement, the residual is:
`r=poses[0].log()`, so I think the jacobian is:
let `t = Log(poses[0])`, 
$ J = \partial Log(poses[0]) / \partial poses[0] = J_r^{-1}(t) $
so this line:
        `J.block<DoF, DoF>(row, col) =  MatrixT::Identity();`
should be:
`J.block<DoF, DoF>(row, col) = poses[0].log().rjacinv();`.
How do you think?

joansola:


examples/se2_sam.cpp
```diff
@@ -451,18 +451,27 @@ int main()
          *  info matrix:
          *     W = I                    // trivial
          *
-         *  residual
-         *     r = W * (poses[0] (-) measurement) = poses[0].log()
+         *  residual uses left-minus since reference measurement is global
+         *     r = W * (poses[0] .- measurement) = log(poses[0] * Id.inv) = poses[0].log()
```
artivis: @joansola How about `.(-)` ??

joansola: Upper in this little doc section I define `.-` as left-minus and `-.` as right minus.

artivis: I saw it, it is just that keeping the parentheses makes it clear at first sight that it is not a regular minus operator. Then it could be any variant `.(-)`, `(.-)`, `'(-)`, `('-)` etc.  
The operator `.-` reminds me of the element-wise operators in e.g. Matlab.

joansola: OK let me take a look
There was a bug in the se3 file anyway. just pushed

joansola: OK I pushed new minus signs as `(.-)` and `(-.)`.

examples/se2_sam.cpp
```diff
@@ -464,14 +464,12 @@ int main()
...
-        r.segment<DoF>(row)         = poses[0].lminus(SE2d::Identity(), J_r_p0).coeffs();
-
-        // Jacobian of residual wrt pose
-        J.block<DoF, DoF>(row, col) = J_r_p0;
+        // residual and Jacobian.
+        // Notes:
+        //   We have residual = expectation - measurement, in global tangent space
+        //   We have the Jacobian in J_r_p0 = J.block<DoF, DoF>(row, col);
+        // We compute the whole in a one-liner:
+        r.segment<DoF>(row)         = poses[0].lminus(SE2d::Identity(), J.block<DoF, DoF>(row, col)).coeffs();
```
artivis: Me gusta one-liner ;P

examples/se2_sam.cpp
artivis: :+1:

examples/se2_sam.cpp
joansola: ;-p

PR #79, opened 2019-05-15
Make very minor stylistic suggestions to the paper.

drvinceknight:
Some very minor stylistic/language suggestions (as part of https://github.com/openjournals/joss-reviews/issues/1371).

PR #81, opened 2019-05-18
Adress #79 about the intro in README.md

joansola:
Remove weird intro paragraph and add a few lines explaining the main assets of manif


artivis:
LGTM, however I think it still misses a couple sentence as why Lie theory and why manif.

Issue #94, opened 2019-08-04
Normalization issue for SO3 with multiple rplus operations


MichalNowicki:
Thanks for this work.

We are considering using _manif_ in our EKF implementation that constantly updates the rotation estimate (SO3) with angular velocity updates stored in Lie algebra. It is working on our own simple LieAlgebra library but it crashes after several iterations when using _manif_. Minimal code that reproduces the error:
```
 // Identity
manif::SO3d R = manif::SO3d(0, 0, 0, 1.0);

// Some angular velocity
manif::SO3Tangentd Om = Eigen::Vector3d(0.001, 0.0, 0.0);

// New orientation
for (int i = 0 ; i < 1000; i++) {
   std::cout << i << std::endl;
   R = R.plus(Om);
}
```
I guess it is related to pull request https://github.com/artivis/manif/pull/72 and I understand your arguments but as for now I think it is necessary to perform R.normalize() after every change to make sure it is still a proper quaternion. Maybe it should be added to the examples concerning EKF/optimization?

MichalNowicki:
Corrected code if someone encounters the same issue:
```
for (int i = 0 ; i < 1000; i++) {
   R = R.plus(Om);
   R.normalize();
}
```

joansola:
Hi Michal, thanks for pointing this out.

It is strange. Some time ago, I was trying in Eigen to produce the same error. I did one million iterations of quaternion composition and regarded the norm of the result at the end. It was 1.000000 and a lot of zeros.

Your experience with much fewer iterations makes me re-consider what I did. I will come later with more comments. Since we re-use quite a lot of Eigen algebra inside manif, the following comes to mind:

  - maybe our exp() function is wrong;
  - maybe the small-angle approximations are too severs -- though you use 0.001 which is not that small.
  - maybe my experiment and yours are not equivalent in some sense.

MichalNowicki:
For now, from my mobile - it was constructor message about norm not being 1.0. From my debugging, it is almost 1.0 but check is very strict with eps 1e-15. I can post exact message later

joansola:
The code:
```c++
    // Identity
   manif::SO3d R = manif::SO3d(0, 0, 0, 1.0);

   // Some angular velocity
   manif::SO3Tangentd Om = Eigen::Vector3d(0.001, 0.0, 0.0);

   // New orientation
   for (int i = 0 ; i < 1000; i++) {
      R = R.plus(Om);
[...]

PR #95, opened 2019-08-04
Address #94: Add test for norm stability

joansola:
See #94 

Repeated exp and compose bring the quaternion out of unit norm. This is a numerical issue, and it is normal.

The thing is maybe we are being too strict in the `SO3` constructor and when the squaredNorm of the quaternion is off by more than 1e-15 (`Constants::eps_s`) we throw.

A suggestion is to relax the threshold to some 1e-10 --> use `Constants::eps` instead.

### DIFF
--- a/test/so3/gtest_so3.cpp
+++ b/test/so3/gtest_so3.cpp
@@ -648,6 +648,21 @@ TEST(TEST_SO3, TEST_SO3_RIGHT_LEFT_JAC)
   EXPECT_EIGEN_NEAR(tan.rjac(), tan.ljac().transpose());
 }
 
+TEST(TEST_SO3, TEST_SO3_NORM_STABILITY)
+{
+    // Identity
+   manif::SO3d R = manif::SO3d(0, 0, 0, 1.0);
+
+   // Some angular velocity
+   manif::SO3Tangentd Om = Eigen::Vector3d(0.001, 0.0, 0.0);
+
+   // New orientation
+   for (int i = 0 ; i < 1000; i++) {
+      R = R.plus(Om);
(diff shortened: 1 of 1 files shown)


test/so3/gtest_so3.cpp
```diff
@@ -648,6 +648,21 @@ TEST(TEST_SO3, TEST_SO3_RIGHT_LEFT_JAC)
...
+TEST(TEST_SO3, TEST_SO3_NORM_STABILITY)
+{
+    // Identity
+   manif::SO3d R = manif::SO3d(0, 0, 0, 1.0);
+
+   // Some angular velocity
+   manif::SO3Tangentd Om = Eigen::Vector3d(0.001, 0.0, 0.0);
+
+   // New orientation
+   for (int i = 0 ; i < 1000; i++) {
```
artivis: @joansola Please consider adding a [`EXPECT_NO_THROW`](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md) macro around this loop to turn it into a proper unit-test.

artivis:
This is indeed a numerical issue. Notice that probably affects SO(2)/SE(2) as well.  
I think the proper fix would be to test for unit-quaternion (/complex) after exp/compose and re-normalize if necessary.  
@joansola We should turn this small test into a generic one, as in a member of [`CommonTester`](https://github.com/artivis/manif/blob/devel/test/common_tester.h#L110).

joansola:
Yes, the test was only to have a record. Before this PR it needs to get finished. I don't know if you want it where I put it or anywhere else.

artivis:
@joansola I'll add a generic test to replace this one :+1:.

artivis:
Close this in favor of #97.

PR #97, opened 2019-08-04
Fix numerical instability in composition

### LINKED ISSUE #94: Normalization issue for SO3 with multiple rplus operations

artivis:
In `compose`, check for numerical drift in complex/quaternion norm and normalize if necessary.

Use an approximation of `1/sqrt(x)` at `x=1` (see [here](https://github.com/artivis/manif/compare/fix/numerical_stability?expand=1#diff-8bde907ebb733ec2326c6d4fde67db7aR50)) to avoid 'expensive' `sqrt` call.

Fix #94.
Replace #95.

### DIFF
--- a/include/manif/impl/so3/SO3_base.h
+++ b/include/manif/impl/so3/SO3_base.h
@@ -248,6 +248,8 @@ SO3Base<_Derived>::compose(
     OptJacobianRef J_mc_ma,
     OptJacobianRef J_mc_mb) const
 {
+  using std::abs;
+
   static_assert(
     std::is_base_of<SO3Base<_DerivedOther>, _DerivedOther>::value,
     "Argument does not inherit from S03Base !");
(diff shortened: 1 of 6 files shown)


include/manif/impl/so3/SO3_base.h
```diff
@@ -262,7 +264,17 @@ SO3Base<_Derived>::compose(
   if (J_mc_mb)
     J_mc_mb->setIdentity();
 
-  return LieGroup(quat() * m_SO3.quat());
+  QuaternionDataType ret_q = quat() * m_SO3.quat();
+
+  const Scalar ret_sqnorm = ret_q.norm();
```
joansola: Should be `ret_q.squaredNorm()` no?

artivis:
Yes SE(3) relies on SO(3) and thus is addressed automatically.

Good catch for the 'norm', I'll fix it asap.

artivis:
@joansola Fixed in SO3 `norm` -> `squaredNorm`

PR #102, opened 2019-08-18
Fix wrong type deduction in test

artivis:
The function `toRad` is template en deduces the input argument type. In a bunch of so3 tests it deduces type `int` which in turn leads to severe precision loss.  
Fixing the type deduction uncovered some handcrafted failing so3 tests. Not sure yet if those tests are simply wrong or else if there is an actual bug to fix.

The tests need to be fixed before merging this.

### DIFF
--- a/test/so3/gtest_so3.cpp
+++ b/test/so3/gtest_so3.cpp
@@ -476,10 +476,10 @@ TEST(TEST_SO3, TEST_SO3_BETWEEN)
 
   auto so3c = so3a.between(so3b);
 
-  EXPECT_DOUBLE_EQ(0, so3c.x());
-  EXPECT_DOUBLE_EQ(0, so3c.y());
-  EXPECT_DOUBLE_EQ(0, so3c.z());
-  EXPECT_DOUBLE_EQ(1, so3c.w());
+  EXPECT_NEAR(0, so3c.x(), 1e-15);
+  EXPECT_NEAR(0, so3c.y(), 1e-15);
+  EXPECT_NEAR(0, so3c.z(), 1e-15);
+  EXPECT_NEAR(1, so3c.w(), 1e-15);
 }
 
 /// with Jacs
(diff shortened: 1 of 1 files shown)


artivis:
@joansola If you could help out on this issue.

artivis:
> Where is this fcn toRad() comeing from?

It is in `manif/impl/utils.h`.

joansola:
I see. Then I guess this story of just adding a dot after the numbers is what must be done. Otherwise cast the numbers to double. 

But your description points more issues to be taken care of, related to failing tests. What about these thests?

artivis:
@joansola If you pull/compile/test this branch you will see that a bunch of so3 tests are now failing. Likely those for which `toRad` was fixed.
Those tests are 'handwritten', they are not part of the generic common tests. I overlooked it but I believe the generic common tests are still passing fine.
My guess is that some of the failing tests are simply wrong, others are simply failing on precision.

joansola:
And why should this test give identity?
```c++
TEST(TEST_SO3, TEST_SO3_COMPOSE)
{
    // compose two particular elements giving identity as result
  SO3d so3a(toRad(-165.),toRad(-135.),toRad(-90.));
  SO3d so3b(toRad(15.),toRad(45.),toRad(90.));

  auto so3c = so3a.compose(so3b);

  EXPECT_DOUBLE_EQ(0, so3c.x());
  EXPECT_DOUBLE_EQ(0, so3c.y());
  EXPECT_DOUBLE_EQ(0, so3c.z());
  EXPECT_DOUBLE_EQ(1, so3c.w());
}
```
It is failing but I don't see why it should result in the identity. There are many other similar tests.

For example this other one:
```c++
TEST(TEST_SO3, TEST_SO3_BETWEEN)
{
    // between two equals is identity
  SO3d so3b(toRad(15.),toRad(45.),toRad(90.));
  SO3d so3a(toRad(-15.),toRad(-45.),toRad(-90.));

  auto so3c = so3a.between(so3b);

  EXPECT_NEAR(0, so3c.x(), 1e-15);
  EXPECT_NEAR(0, so3c.y(), 1e-15);
  EXPECT_NEAR(0, so3c.z(), 1e-15);
  EXPECT_NEAR(1, so3c.w(), 1e-15);
}
```
you say two equals, but clearly they are not two equals since one is the negative of the other. What's the point with these tests???

artivis:
Those tests are wrong, that's my point. So either we fix them or remove them altogether since those features are tested by the generic set of tests.

joansola:
Some can be converted to canonical tests, but maybe they are covered in the macro tests. For example:

compose(1,q) == q
compose(q,1) == q
compose (q,q*) == 1
compose (q*,q) == 1
between(q,q) == 1

These , however, apply to all Lie groups and therefore should be included as part of the standard test packages in the macros (with q* replaced by q.inverse()). Are they?

joansola:
I see COMPOSE_WITH_INV and BETWEEN_SELF.

I do not see COMPOSE_LIE_WITH_ID nor COMPOSE_ID_WITH_LIE  or the like. These should be added to the standard test suite.

joansola:
As per these SO3 tests, they can be just removed!!

artivis:
> I do not see COMPOSE_LIE_WITH_ID nor COMPOSE_ID_WITH_LIE or the like. These should be added to the standard test suite.

There are tested [here](https://github.com/artivis/manif/blob/devel/test/common_tester.h#L233), they simply didn't get their own function/macro.

joansola:
OK then I think it is wise to remove all those weird tests.
 I'll approve this review when this is done.

artivis:
Closing this PR in favor of a clean slate in #103.

PR #104, opened 2019-08-19
SAM 3D with self-calibration

joansola:
This PR adds a new example of a Smoothing and Mapping problem with self-calibration of the motion sensor parameters.

It follows the example in the paper, section V.C, but in 3D.


artivis:
@joansola pushed a few small fixes, in particular I saw an opportunity to highlight the use of `Eigen::Map` (see 95166f0).

PR #109, opened 2019-11-14
Add trivial groups Rn

### LINKED ISSUE #108: Trivial euclidean manifold E(n) ? (feature request)
Dear all,

I'm an application which takes an "manif" manifold as a template parameter. This works nicely, since the basic interface, accessors and methods are naturally the same for all manifolds.

However, it fails as soon as I want to operate in Euclidean space. What is missing, is essentially a manif-implementation of the must fundamental n-dimensional Euclidean manifold E(n), in which case anything within manif should fall back to the natural Eigen-types and methods.

Is there any plan to add such a feature?

Best, and thanks in advance!

artivis:
Add the trivial groups Rn.

Close #108.


artivis:
@joansola This PR is not fully finished yet but ready for a first round of review.

artivis:
@markusgft FYI.

include/manif/impl/rn/Rn_base.h
```diff
@@ -0,0 +1,226 @@
...
+  T.template topRightCorner<Dim,1>() = coeffs();
+  return T;
+}
+
+template <typename _Derived>
+typename RnBase<_Derived>::LieGroup
+RnBase<_Derived>::inverse(OptJacobianRef J_minv_m) const
+{
+  if (J_minv_m)
+    J_minv_m->setIdentity()*=Scalar(-1);
```
joansola: I would like the *= to come separated by pre- and post- spaces... just for reading.

Blame me for this one!

include/manif/impl/rn/Rn_base.h
```diff
@@ -0,0 +1,226 @@
...
+};
+
+template <typename _Derived>
+typename RnBase<_Derived>::Transformation
+RnBase<_Derived>::transform() const
+{
+  Transformation T(Transformation::Identity());
+  T.template topRightCorner<Dim,1>() = coeffs();
+  return T;
+}
```
joansola: Unsure of this one. What is the reason to have to use this .topRightCorner thing?

include/manif/impl/rn/Rn_base.h
```diff
@@ -0,0 +1,226 @@
...
+  static_assert(
+    RnBase<_DerivedOther>::Dim==_DerivedOther::Dim, "Dimension mismatch !");
+
+  if (J_mc_ma)
+    J_mc_ma->setIdentity();
+
+  if (J_mc_mb)
+    J_mc_mb->setIdentity();
+
+  return LieGroup(coeffs() + m.coeffs());
```
joansola: OK

include/manif/impl/rn/Rn_base.h
```diff
@@ -0,0 +1,226 @@
...
+  {
+    J_vout_m->setIdentity();
+  }
+
+  if (J_vout_v)
+  {
+    J_vout_v->setIdentity();
+  }
+
+  return v + coeffs();
```
joansola: I know it's stupid and commutative, but I would have written here `coeffs() + v`;

include/manif/impl/se2/SE2.h
```diff
@@ -61,6 +61,8 @@ struct SE2 : SE2Base<SE2<_Scalar>>
   MANIF_COMPLETE_GROUP_TYPEDEF
   using Translation = typename Base::Translation;
   MANIF_INHERIT_GROUP_API
+  using Base::transform;
+  using Base::rotation;
```
joansola: Unrelated to Rn thing no? Do you want this on the same PR?

test/common_tester.h
```diff
@@ -8,9 +8,7 @@
 #define MANIF_TEST(manifold)                                              \
```
joansola: Did you run all the standard tests on Rn?

include/manif/impl/rn/Rn.h
```diff
@@ -0,0 +1,162 @@
...
+};
+
+template <typename _Scalar> using R2 = Rn<_Scalar, 2>;
+template <typename _Scalar> using R3 = Rn<_Scalar, 3>;
+template <typename _Scalar> using R4 = Rn<_Scalar, 4>;
+template <typename _Scalar> using R5 = Rn<_Scalar, 5>;
+template <typename _Scalar> using R6 = Rn<_Scalar, 6>;
+template <typename _Scalar> using R7 = Rn<_Scalar, 7>;
+template <typename _Scalar> using R8 = Rn<_Scalar, 8>;
+template <typename _Scalar> using R9 = Rn<_Scalar, 9>;
```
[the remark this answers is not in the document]
joansola: This is still unsure , but anyway, why not R1!

include/manif/impl/rn/Rn.h
```diff
@@ -0,0 +1,162 @@
...
+  template <typename _DerivedOther>
+  Rn(const LieGroupBase<_DerivedOther>& o);
+
+  // Copy constructor given Eigen
+  template <typename _EigenDerived>
+  Rn(const Eigen::MatrixBase<_EigenDerived>& data);
+
+  // LieGroup common API
+
+  //! Get a const reference to the underlying DataType.
```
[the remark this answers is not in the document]
artivis: That's only to mark documentation.

include/manif/impl/rn/Rn.h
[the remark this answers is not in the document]
artivis: Right, I'll add it.

include/manif/impl/rn/RnTangent_base.h
```diff
@@ -0,0 +1,201 @@
...
+  return ljac();
+}
+
+template <typename _Derived>
+typename RnTangentBase<_Derived>::Jacobian
+RnTangentBase<_Derived>::smallAdj() const
+{
+  static const Jacobian smallAdj = Jacobian::Identity();
+  return smallAdj;
+}
```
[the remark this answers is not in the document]
artivis: No I actually do not know. It's just a temporary implementation until it is figured out.

include/manif/impl/rn/Rn_base.h
artivis: `topRightCorner` is just another way to access the top-right block of the matrix, here essentially the most part of the right column.

test/common_tester.h
artivis: Yes, the tests do pass locally and on Travis CI. See the tests [here](https://github.com/artivis/manif/blob/feature/rn/test/rn/gtest_rn.cpp#L15) and Travis CI [here](https://travis-ci.com/artivis/manif/builds/136435353).

include/manif/impl/se2/SE2.h
artivis: I had to remove those from a macro that is meant to apply to all groups but Rn does not have `rotation` of course. So yeah it belongs to the same load of changes to get Rn.

include/manif/impl/rn/RnTangent_base.h
```diff
@@ -0,0 +1,201 @@
...
+
+template <typename _Derived>
+typename RnTangentBase<_Derived>::LieAlg
+RnTangentBase<_Derived>::hat() const
+{
+  LieAlg t_hat = LieAlg::Constant(0);
+  t_hat.template topRightCorner<Dim, 1>() = coeffs();
+  return t_hat;
+}
```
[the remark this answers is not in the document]
artivis: This essentially implement `t^` from Appendix E.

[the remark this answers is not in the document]
joansola: Oh, OK, sure. The topRightCorner think confuses me.

include/manif/impl/rn/RnTangent_base.h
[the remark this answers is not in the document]
joansola: OK let us keep an eye here, this is a bug factory!

include/manif/impl/rn/Rn_base.h
joansola: Ahhh it's a transformation matrix only fro translation, so `T = [ I t ; 0 1 ]`

OK

joansola:
OK good job. We have a warning TODO on the `smallAdj`, I just put it here for highlighting it.

joansola:
A side, general question: Why Rn and not e.g. Tn? In the paper I wrote T(n) for these groups. T is for 'translation'.

This comment is no big deal. But it might be interesting to come up with a name that the community can feel right.

artivis:
> OK good job. We have a warning TODO on the `smallAdj`, I just put it here for highlighting it.

Yeah I could use the macro that throws a 'Not implemented yet' message for safety.
Also we should implement a generic test for `smallAdj`, I couldn't find any after a quick search.

artivis:
> A side, general question: Why Rn and not e.g. Tn? In the paper I wrote T(n) for these groups. T is for 'translation'.

I used `Rn` because you seemed to favor it in #108 discussion. I don't mind renaming it (once) so lets agree on what it should be.

joansola:
> > OK good job. We have a warning TODO on the `smallAdj`, I just put it here for highlighting it.
> 
> Yeah I could use the macro that throws a 'Not implemented yet' message for safety.
> Also we should implement a generic test for `smallAdj`, I couldn't find any after a quick search.

I know of some properties of smallAdj but I am afraid they only hold for multiplicative groups:

for x,y \in Tangent, then

adj_x * y = vee( x*y - y*x)

For additive groups this may collapse to a trivial thing, since x*y = x+y and then the whole thing goes to zero -- does this mean that smallAdj = 0?

This is the mistery and the danger by now.

joansola:
> > A side, general question: Why Rn and not e.g. Tn? In the paper I wrote T(n) for these groups. T is for 'translation'.
> 
> I used `Rn` because you seemed to favor it in #108 discussion. I don't mind renaming it (once) so lets agree on what it should be.

I tend to favor T(n). But unlike SO(n) which are well established, and also S^n, the name T(n) I just invented it from thin air in the paper. I have seen T(n) to signify the torus of dimension n, which is another thing.

artivis:
> I tend to favor T(n). But unlike SO(n) which are well established, and also S^n, the name T(n) I just invented it from thin air in the paper. I have seen T(n) to signify the torus of dimension n, which is another thing.

Ok then we have to be careful, maybe Tn is not the best solution.

joansola:
> > I tend to favor T(n). But unlike SO(n) which are well established, and also S^n, the name T(n) I just invented it from thin air in the paper. I have seen T(n) to signify the torus of dimension n, which is another thing.
> 
> Ok then we have to be careful, maybe Tn is not the best solution.

From Wikipedia https://en.wikipedia.org/wiki/Translation_(geometry)

    Seems that T(n) is a good name.

From wikipedia https://en.wikipedia.org/wiki/Euclidean_group

    The E(n) group is the Euclidean group, and it is not the same because Euclidean group includes rotation.

From StackExchange https://math.stackexchange.com/questions/666870/generators-of-translation-lie-algebra 

    the group essentially R^n , which is an additive group isomorphic to the matrix group of translations.

I feel going for T(n) is safe. I do not like R^n because it is not a group, it is a set, or a space, and it lacks the group law of composition. You could say that T(n) = {R(n), +}.

artivis:
> for x,y \in Tangent, then
> 
> adj_x * y = vee( xy - yx)

Did you mean,

`adj_x*y^ = x^*y^ − y^*x^` ? 
That's the Lie bracket `[x^,y^]` right ?

artivis:
> I feel going for T(n) is safe. I do not like R^n because it is not a group, it is a set, or a space, and it lacks the group law of composition. You could say that T(n) = {R(n), +}.

Ok then I'll rename to `Tn` in a bit, in case you change your mind ;p

joansola:
> > for x,y \in Tangent, then
> > adj_x * y = vee( xy - yx)
> 
> Did you mean,
> 
> `adj_x*y^ = x^*y^ − y^*x^` ?
> That's the Lie bracket `[x^,y^]` right ?

Yes and no. We are mixing lie algebra and vector representations of the tangent elements x and y. Also, the small adjoint that we have is in reality the small adjoint matrix. Here are the equivalences:

Given:
  - Vectors `x`, `y` in tangent space
  - Elements `x^`, `y^` in Lie algebra
  - Small adjoint operator `adj_x`
  - Small adjoint matrix `[adj_x]`

In Lie algebra, we have

    adj_x(y) = x^*y^ - y^*x^

In vector space we have

    [adj_x] * y = ( x^*y^ - y^*x^ ) v

where 'v' is the vee operator.

But, the thing is, in additive groups like the one we are handling, we have `x^=x` and `*=+`, so the Lie bracket resolves to zero:

    (x+y) - (y-x) = 0

and so I am a little confused as to what is the small adjoint in such case. Following this reasoning, it should be simply `[adj_x] = 0` !!

artivis:
Oh yeah I had forgot that distinction `adj` / `[adj]`...
Ok then the only issue with implementing `[adj_x] * y = ( x^*y^ - y^*x^ ) v` is that we don't have the `v` operator implemented. 
I'll add it to the todo list...

joansola:
BTW, one way to see this is the following:

Consider T(n) as the subgroup of SE(n) with null rotation (that is, R = I).

Then, the adjoint of T(n) and the adjoint of SE(n) should be equal if we set theta=0 in the adjoint of SE(n).

Of course the adjoint of SE(n) has a larger dimension. We must look at the translation part only.

But if it happens that in SE(n) we have adj = 0 when theta=0, then we can conclude that the adjoint of T(n) is also zero.

joansola:
> > I feel going for T(n) is safe. I do not like R^n because it is not a group, it is a set, or a space, and it lacks the group law of composition. You could say that T(n) = {R(n), +}.
> 
> Ok then I'll rename to `Tn` in a bit, in case you change your mind ;p

According to wikipedia https://en.wikipedia.org/wiki/Table_of_Lie_groups

The group name is finally R^n !!!!

:-)

joansola:
All my research leads to smallAdj = 0_{nxn}. But let us not claim victory yet.

artivis:
> According to wikipedia https://en.wikipedia.org/wiki/Table_of_Lie_groups 
> The group name is finally R^n !!!!

Great, less work for me :)

joansola:
> All my research leads to smallAdj = 0_{nxn}. But let us not claim victory yet.

Another way to see it is through the formula given by Eade and Fourmy etal

`J_r(\bftau)  = \sum_i \frac{(-\ad{\bftau})^i}{(i+1)!}`

which is an infinite sum of powers of `adj`, starting with `adj^0=I`. Since `J_r = I`, we have that `adj` must be zero!

joansola:
OK Just for the fun of it. Another way to show that the small adjoint is zero.

Take the matrix form of the group. For translations `t` and `s`, their Lie algebras are `tau = [ 0 t ; 0 0 ]` and `sigma = [ 0 s ; 0 0 ]`.

Now the small adjoint operator is

`adj_tau (sigma) = tau*sigma - sigma*tau = [0 0;0 0] - [0 0;0 0] = [0 0;0 0]`

which is a perfect zero. The matrix version is consequently zero too:

`[adj_t] * s = [0 0;0 0]v = [0;0]` --> `[adj_t] = [0 0;0 0]` voila.

test/common_tester.h
joansola: Does your answer still apply after further commits?

include/manif/impl/rn/RnTangent_base.h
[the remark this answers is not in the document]
joansola: Solved. closing.

test/common_tester.h
artivis: Yes, only the documentation-related test is failing. I'll look it up.

artivis:
CI is happy now!
I only have to update the doc and the readme then this is good to be merged.

joansola:
@artivis FYI

```terminal
[==========] 828 tests from 36 test suites ran. (719 ms total)
[  PASSED  ] 825 tests.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_RMINUS_JACOBIANS
[  FAILED  ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_LMINUS_JACOBIANS
[  FAILED  ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_MINUS_JACOBIANS

 3 FAILED TESTS
```

full log of failed tests:

```terminal
[ RUN      ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_RMINUS_JACOBIANS
/Users/jsola/dev/manif/test/rn/../common_tester.h:727: Failure
Value of: manif::isManifNear(state_pert, state_lin, "state_pert", "state_lin", tol_)
  Actual: false (state_pert != state_lin
state_pert:
0.011310875415802002
state_lin:
0.011310858651995659
diff:
1.6763806343078613e-08
)
Expected: true
/Users/jsola/dev/manif/test/rn/../common_tester.h:734: Failure
Value of: manif::isManifNear(state_pert, state_lin, "state_pert", "state_lin", tol_)
  Actual: false (state_pert != state_lin
state_pert:
0.011417090892791748
state_lin:
0.011417107656598091
diff:
-1.6763806343078613e-08
)
Expected: true
[  FAILED  ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_RMINUS_JACOBIANS (1 ms)
[ RUN      ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_LMINUS_JACOBIANS
/Users/jsola/dev/manif/test/rn/../common_tester.h:747: Failure
Value of: manif::isManifNear(state_pert, state_lin, "state_pert", "state_lin", tol_)
  Actual: false (state_pert != state_lin
state_pert:
0.011310875415802002
state_lin:
0.011310858651995659
diff:
1.6763806343078613e-08
)
Expected: true
/Users/jsola/dev/manif/test/rn/../common_tester.h:754: Failure
Value of: manif::isManifNear(state_pert, state_lin, "state_pert", "state_lin", tol_)
  Actual: false (state_pert != state_lin
state_pert:
0.011417090892791748
state_lin:
0.011417107656598091
diff:
-1.6763806343078613e-08
)
Expected: true
[  FAILED  ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_LMINUS_JACOBIANS (0 ms)
[ RUN      ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_MINUS_JACOBIANS
/Users/jsola/dev/manif/test/rn/../common_tester.h:767: Failure
Value of: manif::isManifNear(state_pert, state_lin, "state_pert", "state_lin", tol_)
  Actual: false (state_pert != state_lin
state_pert:
0.011310875415802002
state_lin:
0.011310858651995659
diff:
1.6763806343078613e-08
)
Expected: true
/Users/jsola/dev/manif/test/rn/../common_tester.h:774: Failure
Value of: manif::isManifNear(state_pert, state_lin, "state_pert", "state_lin", tol_)
  Actual: false (state_pert != state_lin
state_pert:
0.011417090892791748
state_lin:
0.011417107656598091
diff:
-1.6763806343078613e-08
)
Expected: true
[  FAILED  ] TEST_R1f_JACOBIANS_TESTER.TEST_R1f_MINUS_JACOBIANS (0 ms)
```

joansola:
BTW the failed tests above were on my Mac OSX. Maybe on Linux the error is slightly smaller and the tests pass. But see my comments above: the test should pass with error zero!!!

joansola:
> FYI I'm experiencing some test regression on my machine. Some SO3 test are failing due to precision, some other results (sometime) in segfault.
> I'm not sure what has changed, possibly my compiler version, I'll have to check. While investigating I got concerned that we possibly have some memory alignment issues lying around. More investigation required...

No idea what can cause this test regression!

Maybe investigating the Rn tests (comments above) we can find a common cause for all these weakly passing tests?

artivis:
@joansola After a brief investigation, this issue seems to be due to single floating point precision.
See e.g. [this article](https://www.geeksforgeeks.org/g-fact-39-floating-point-operations-associativity/):

> floating point arithmetic may not follow the law of associativity in every case. This is due to the format in which the floating point numbers are stored and represented, it rounds off the numbers during calculations, hence, the associative laws of algebra do not necessarily hold for floating-point numbers.

For short, the order of operations matters; in our case we have:
```
-b+a+w == a+w-b
```
which, as per the above quote, may not always be true...
As a minimal reproducible example please consider [this snippet](https://ideone.com/kf35R8
).

Notice that all failing tests that you linked concern `R1f`.

As a fix we can simply lower the test tolerance.

joansola:
> @joansola After a brief investigation, this issue seems to be due to single floating point precision.
> See e.g. [this article](https://www.geeksforgeeks.org/g-fact-39-floating-point-operations-associativity/):
> 
> > floating point arithmetic may not follow the law of associativity in every case. This is due to the format in which the floating point numbers are stored and represented, it rounds off the numbers during calculations, hence, the associative laws of algebra do not necessarily hold for floating-point numbers.
> 
> For short, the order of operations matters; in our case we have:
> 
> ```
> -b+a+1*w == a+w-b
> ```
> 
> which, as per the above quote, may not always be true...
> As a minimal reproducible example please consider [this snippet](https://ideone.com/kf35R8).
> 
> Notice that all failing tests that you linked concern `R1f`.
> 
> As a fix we can simply lower the test tolerance.

Yes OK, but I think 1e-8 tolerance cannot be explained by just floating point non-commutativity. I would agree with the argument at 1e-20 or beyond, but not at 1e-8 really. Maybe I am wrong.

Can we just try this:

```c++
double a,b,c;
a = random; b = random ; c=random;
double err = ((a + b) + c) - ((a + c) + b);
```
and see what `err` looks like?

The thing is: OK, 1e-8 is small for us. But my point was knowing whether we were doing something wrong. I think it is worth investigating.

artivis:
> Can we just try this:
> 
> ```c++
> double a,b,c;
> a = random; b = random ; c=random;
> double err = ((a + b) + c) - ((a + c) + b);
> ```
> 
> and see what `err` looks like?
> 
> The thing is: OK, 1e-8 is small for us. But my point was knowing whether we were doing something wrong. I think it is worth investigating.

Please have a look at [the snippet I linked previously](https://ideone.com/kf35R8).

joansola:
Oh I see. And yes, it's for floats, right? 

It's kinda disappointing, but OK I see your point. I agree we can ignore this, just lower the tolerance for tests with float -- or for all tests. Maybe there is an `eps` related to the scalar type that you can use to define a type-dependent tolerance?

artivis:
Using a different tolerance for float and double is already been taken care of [here](https://github.com/artivis/manif/blob/feature/rn/test/common_tester.h#L929).
I'll lower the tolerance for float.

artivis:
Doc is updated, this PR is ready.

joansola:
Maybe not immediately, but we'll have to add the actions of `SO(n)` and `SE(n)` on `Rn`. Currently, these actions are performed on regular `Eigen` vectors. Now we'll need to add `Rn` to the API. Of course it's just overloads of the existing implementations.

artivis:
I'll definitely leave this to a following PR. Especially because I still want to rethink/refactor `act` so that we can `act` at least on both points and vectors. I need it in my clean up of the 'tesc' planner code.
But let us reserve this discussion for another thread.

joansola:
Sure. As you prefer!

joansola:
> Also we should implement a generic test for `smallAdj`, I couldn't find any after a quick search.

Did you make this test generic? The test is as follows:

```c++
Tangent x, y;
x = random(); y = random();
x.smallAdj() * y = (x.hat() * y.hat() - y.hat() * x.hat() ).vee();
```
or, if we do not have `vee()`,
```c++
(x.smallAdj() * y).hat() = x.hat() * y.hat() - y.hat() * x.hat() ;
```

artivis:
> Did you make this test generic? The test is as follows:
> 
> ```c++
> Tangent x, y;
> x = random(); y = random();
> x.smallAdj() * y = (x.hat() * y.hat() - y.hat() * x.hat() ).vee();
> ```
> 
> or, if we do not have `vee()`,
> 
> ```c++
> (x.smallAdj() * y).hat() = x.hat() * y.hat() - y.hat() * x.hat() ;
> ```

See #114.

joansola:
What's the status of this PR?

I have one only concern: that `hat()` and `vee()` transform to the Lie algebra of the matrix translation group `T(n)`, that is `x.hat = [ 0 x ; 0 0]`. 

I guess we can live with it, but strictly speaking in the translation vector group `Rn` we have `x.vee()  = x`, which continues to be a vector of R^n...

artivis:
> What's the status of this PR?

I just replaced `x.hat()` so that `x.hat()=x`.
@joansola If you agree, I will merge it once the tests are passed.

markusgft:
Thanks again @artivis and @joansola  for taking action so quickly, I really appreciate it!

joansola:
@artivis but then how about the test of `smallAdj`?

```c++
(x.smallAdj() * y).hat() = x.hat() * y.hat() - y.hat() * x.hat() ;
```

what is the `*` operator doing in this test? Is it group composition, that is, `+` for Rn? 

Solving this consistently is not easy. The `x.hat()` is not a member of the group, but of the Lie algebra, and therefore the group composition operator does not apply to them.

On the other hand, if `*` is matrix multiplication, then `x.hat()` is not a matrix but a vector.

Finally, if we make `x.hat()` a matrix we go back to where we started.

The only way out I suppose is to have the test redefined as so:

```c++
x.smallAdj() * y = x.bracket(y)
```

For matrix Lie groups (and maybe for many multiplicative groups), the bracket is defined as 
```c++
x.bracket(y) = x*y - y*x
```

But this cannot be the case for additive groups such as Rn. The Lie bracket in Rn is zero,
```c++
x.bracket(y) = 0
```

This is why I said that we could 'live' with `x.hat() = [x 0 ; 0 0]`, to save the form of the current test using the multiplicative form of the lie bracket.

All thhese changes could make the point for a new issue and a new PR. But for the time being, defining `x.hat() = x` just breaks down the test on `smallAdj`.

CONCLUSION: I cannot approve this PR as it is now.

markusgft:
Is there roadmap for getting this onto master, @artivis @joansola ?

artivis:
@markusgft, the plan is to wait for #30 to land before merging this. This should happen sooner than later.
The reason is that, if it all goes as planned, #30 will be accompanied by a first tagged release of `manif` and a DOI to ease referencing. #30 has essentially been reviewed in the current state of `devel` and the addition of this PR would probably require a new review, further delaying #30. The process is in its final step, you can follow it [here](https://github.com/openjournals/joss-reviews/issues/1371).
Sorry for the inconvenience.

minrui-hust:
In `RnTangent.h` line 180:
```c++
Ei(i, RnTangentBase<Derived>::DoF) = 1;
```
Ei is a column vector, the second index **RnTangentBase<Derived>::DoF** out of bound.
Test fail and throw an exception, on ubuntu 16.04.

artivis:
> In `RnTangent.h` line 180:
> 
> ```c++
> Ei(i, RnTangentBase<Derived>::DoF) = 1;
> ```
> 
> Ei is a column vector, the second index **RnTangentBase::DoF** out of bound.
> Test fail and throw an exception, on ubuntu 16.04.

Hi @minrui-hust,
`Ei` is a matrix (defined [here](https://github.com/artivis/manif/blob/feature/rn/include/manif/impl/rn/RnTangent.h#L29)) and tests are passing both on Travis and on my machine. Could you provide some more information concerning the exception you encountered??

artivis:
@joansola with #30 being merged, we can now consider merging this PR. Could you give it a last review and eventually approve it?

artivis:
> Did you address the `x.hat()` definition as we spoke? This PR is old and I lost track of all issues, but I do remember that `x.hat()` needed attention. I have seen a commit about this being reverted today, [fc52aa3](https://github.com/artivis/manif/commit/fc52aa3cf1dd0cb44d272dcc937ff90ac9bd138c), and I am a little confused of what has been going on regarding this.

I did revert the change to `x.hat()` so that *now* it returns the Lie algebra of the matrix translation group `T(n)`, that is `x.hat = [ 0 x ; 0 0]`. See implementation [here](https://github.com/artivis/manif/pull/109/files#diff-ad1af0d9b02c017ca9918173b5d9f103R116). 
We agreed that "we can live with it" for now and revisit it if necessary.

artivis:
I cleaned up the commit history hence the force-push.
Merging!

PR #118, opened 2020-01-21
Fix Vector typedef and add small test

### LINKED ISSUE #117: What does 'Dim' mean?
In struct traits<SO2<_Scalar>> in `SO2.h`
```c++
...
using Vector  = Eigen::Matrix<Scalar, DoF, 1>;
```
Should the 'DoF' be 'Dim'?
The DoF of SO2 is 1, but the dimension of the Vector SO2 act on is 2.

artivis:
Fix the `Vector` typedef in all groups.
Add a test that rely on `act` which actually acts on points, not vectors, thus the `Point` typedef in the test.

fix #117

### DIFF
--- a/include/manif/impl/so2/SO2.h
+++ b/include/manif/impl/so2/SO2.h
@@ -31,8 +31,8 @@ struct traits<SO2<_Scalar>>
 
   using Jacobian       = Eigen::Matrix<Scalar, DoF, DoF>;
   using Transformation = Eigen::Matrix<Scalar, 3, 3>;
-  using Rotation       = Eigen::Matrix<Scalar, 2, 2>;
-  using Vector         = Eigen::Matrix<Scalar, DoF, 1>;
+  using Rotation       = Eigen::Matrix<Scalar, Dim, Dim>;
+  using Vector         = Eigen::Matrix<Scalar, Dim, 1>;
 };
 
 } /* namespace internal */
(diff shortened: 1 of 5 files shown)


artivis:
This bug&fix also call for moving forward on `act` API so that we can act on different objects (point/vector/etc).

include/manif/impl/so2/SO2.h
```diff
@@ -31,8 +31,8 @@ struct traits<SO2<_Scalar>>
 
   using Jacobian       = Eigen::Matrix<Scalar, DoF, DoF>;
   using Transformation = Eigen::Matrix<Scalar, 3, 3>;
```
joansola: What about <Dim+1,Dim+1> instead of <3,3> ?

include/manif/impl/so3/SO3.h
```diff
@@ -32,7 +32,7 @@ struct traits<SO3<_Scalar>>
   using Jacobian       = Eigen::Matrix<Scalar, DoF, DoF>;
   using Transformation = Eigen::Matrix<Scalar, 4, 4>;
```
joansola: see comment L33

test/common_tester.h
```diff
@@ -547,6 +549,17 @@ class CommonTester : public ::testing::Test
...
+  {
+    using Point = typename LieGroup::Vector;
+
+    Point pin = Point::Random();
+
+    Point pout = LieGroup::Identity().act(pin);
+
+    EXPECT_EIGEN_NEAR(pin, pout);
+  }
+
```
joansola: Why not? But could we imagine a second test a little more conclusive?

I will approve the review anyway regardless of this comment

artivis: Sure we can add a better test later on, for now I just want the `Vector` typedef to be compiled and used to assert its dimensions.
I'll merge and come back to it later.

PR #122, opened 2020-02-04
Do not install package manifest

### LINKED ISSUE #120: manif creates a package.xml file in devel-folder after build
Dear all,

I am using manif in a catkin (ROS) workspace and encounter the following issue:
When building using `catkin build`, manif creates a _package.xml_ file in `<path_to_ws>/devel/share/manif`

Creating a _package.xml_ in the devel folders leads ROS to "believe" that there is a ROS package in devel with the same package name as the orginal package in the src folder. This triggers errors in some ROS tools, in particular `rosdep install`. Normally ROS packages do not have a _package.xml_ in their devel folder after build, which leads me to believe that this is a manif problem.

Is that package.xml in the devel folder used, or maybe just legacy?

artivis:
Do not install the `package.xml` file in `devel/share/manif`.
Fix #120

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,10 +156,6 @@ install(
   DESTINATION include
 )
 
-install(FILES package.xml
-  DESTINATION share/${PROJECT_NAME}
-#  DESTINATION "${config_install_dir}"
-)
 
 ##############
 ## Examples ##


CMakeLists.txt
```diff
@@ -156,10 +156,6 @@ install(
   DESTINATION include
 )
 
-install(FILES package.xml
-  DESTINATION share/${PROJECT_NAME}
```
joansola: Is `share` a global directory, as in `/share`?

artivis: `share` is a relative path here, from cmake doc:
`If a relative path is given it is interpreted relative to the value of the CMAKE_INSTALL_PREFIX variable.`
And,
`[CMAKE_INSTALL_PREFIX] defaults to /usr/local on UNIX and c:/Program Files/${PROJECT_NAME} on Windows.`

joansola: cool thanks

PR #124, opened 2020-02-13
Add default eps to TangentBase::isApprox

### LINKED ISSUE #123: .isApprox() with default tolerance for Tangent-types?
Just out of curiousity: is there a reason why the method .isApprox() in the Tangent Base class does not have a default value for the tolerance?

If not, would it be an option to add a reasonable default tolerance?

artivis:
Add a default `eps` value to `TangentBase::isApprox`.
This should have been done in d584b109e56e68db966e54b98548f5f7317d95f9.

Fix #123.

Edit:
Pushed two extra commits, didn't want to open a separate PR for those:
- download `gtest` `v1.8.x` explicitly instead of `master`. Branch `master` had some recent changes that requiers the `cpp11` flag for older compilers (found that [here](https://github.com/oracle/ktf/issues/10#issuecomment-440629294)).
- download `Eigen` from the official Gitlab repo rather than the deprecated Bitbucket one in CI. There were some spurious timeout issues in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ before_script:
   - cd build
 
   # Update / Install Eigen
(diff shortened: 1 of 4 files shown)


artivis:
@joansola Small review please?

PR #131, opened 2020-05-16
Add theory cheat sheets

artivis:
This PR adds a Lie group theory cheat sheets.


joansola:
* 3-sphere: lacking symbol for the group composition law after S^3 in second column
* S^1, Log: perhaps use arctan2 instead of arctan--> we need 4 quadrants
* Use `^\top` and not `^T` for transposition
* Use "quaternion" and not "3 sphere" as the name of the group, just for kindness to the reader. We used "complex number" for S^1, so it makes sense to use "quaternion" for S^3
* Explicit right-$\oplus$ and right-$\ominus$ in the Jacobians table
* Explicit "Right Jacobians" in the Jacobians table

artivis:
* ~~3-sphere: lacking symbol for the group composition law after S^3 in second column~~
* ~~S^1, Log: perhaps use arctan2 instead of arctan--> we need 4 quadrants
* ~~Use ^\top and not ^T for transposition~~
* ~~Use "quaternion" and not "3 sphere" as the name of the group, just for kindness to the reader. We used "complex number" for S^1, so it makes sense to use "quaternion" for S^3~~
* ~~Explicit right-$\oplus$ and right-$\ominus$ in the Jacobians table~~
* ~~Explicit "Right Jacobians" in the Jacobians table~~
* ~~Skew symmetric matrix is wrong for SO(3). Check out Example 3 on the paper, use that formula.~~
* ~~(I remember I sent you all formulas for the left Jacobians. It might be worth adding them, with a note saying that Manif computes the right, and the formula (57) to convert from right to left Jacobian)~~

I've shared with you the link to the latex source. I'll update the pdf here with the fixes to the above comments.

joansola:
Use also arctan2 for SO2

joansola:
Fixed arctantwo in overleaf

artivis:
Ok, I'll update the pdf here.

joansola:
Updated the comment on left/right jacobians.

The left Jacobians are not the same as "the left Jacobian of a group". This second, in singular, is defined as the "left Jacobian of the Exp() function". This is clear in the paper. The name confusion is not our fault, it's the literature that has it like this.

The same happens with the right Jacobians and "the Right Jacobian of a group".

If you want to add a reference to the paper in the cheat sheet, I think it's good to have because this cheat sheet can become detached from the project if people start sharing it.

artivis:
Updated the document to add a reference and link to the paper.  
@joansola can you have a look before I update the PR?

joansola:
Looks good to me

artivis:
Can you approve the PR so that I merge once the CI passed?

PR #137, opened 2020-06-12
Add public non-const coeffs

### LINKED ISSUE #132: Non-const getter for coeffs()
Dear all,
I am working on an optimal control library which supports both native Eigen-types and manif-types, as well as optimal control on manifolds.
I am trying to make things as generic as possible, i.e. write my code such that Eigen-types and manif-types are interchangeable where possible. 
One of the unfortunate obstacles in the way is that manif does not have a non-const getter for `coeffs()`.
Do you see any major obstacles for adding such a feature?
Any thoughts and discussion on this would be highly appreciated.
Best!

artivis:
Add a public non-const `coeffs` to `LieGroup` objects.

Also remove the protected `coeffs_nonconst`.

Fix #132.

### DIFF
--- a/include/manif/impl/lie_group_base.h
+++ b/include/manif/impl/lie_group_base.h
@@ -36,15 +36,13 @@ struct LieGroupBase
   template <typename _Scalar>
   using LieGroupTemplate = typename internal::traitscast<LieGroup, _Scalar>::cast;
 
-protected:
-
-  //! @brief Access the underlying data by reference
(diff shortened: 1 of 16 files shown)


artivis:
The tests fail because of the newly introduced arm64 job which uses gcc-9 resulting in a bunch of error about deprecated implicit operators.  
I'll merge this since the other tests are passing and open a ticket to support gcc-8/9.

PR #138, opened 2020-06-13
Add MANIF_ASSERT

artivis:
Add macro `MANIF_ASSERT`.  

The macro asserts that a condition is true or throws otherwise. It replaces `MANIF_CHECK` for a couple numerical checks as requested in [this comment](https://github.com/artivis/manif/issues/132#issuecomment-640200391).

`MANIF_ASSERT` macro can be disabled by either defining `MANIF_NO_DEBUG` or the usual `NDEBUG`.

### DIFF
--- a/include/manif/impl/macro.h
+++ b/include/manif/impl/macro.h
@@ -3,6 +3,12 @@
 
 #include <stdexcept> // for std::runtime_error
 
+#ifdef NDEBUG
+# ifndef MANIF_NO_DEBUG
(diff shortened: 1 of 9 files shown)


artivis:
The tests fail because of the newly introduced arm64 job which uses gcc-9 resulting in a bunch of error about deprecated implicit operators.  
I'll merge this as soon as approved since the other tests are passing and open a ticket to support gcc-8/9.

artivis:
@joansola Can you please review/approve?

PR #151, opened 2020-08-02
Cheat sheet small fix

artivis:
Replace constraint `R^T R = I` by `M^{-1} M = I` for `SE2` and `SE3`.


joansola:
Last minute recall: `M.tr * M = I` is actually wrong!

artivis:
ah crap, fixing the fix...

artivis:
Done.

artivis:
For some uploading from the web interface does not seem to work. I'll upload again from the cli asap.

joansola:
I still see `M.tr*M=I`. Valid constraints are either `R.tr*R+I` (precise constraint) or the general `M.inv*M=I` (general abstract constraint). I prefer `R.tr*R+I`.

artivis:
> I prefer `R.tr*R+I`.

You mean `R.tr*R=I`? As it currently is on master? If so let's close this PR and keep the current status.

artivis:
Let's stick with the current status (`R.tr*R=I`) since it seems to be preferred. Closing

Issue #152, opened 2020-08-03
Add SE_2(3) extended pose Lie group manifold to manif


prashanthr05:
First of all, thank you for this great repository! 

I would like to make a feature request for adding an additional Lie group $SE_2(3)$ (extended pose or double direct spatial isometry as introduced in [The invariant extended Kalman filter as a stable observer, appendix A2 ](https://arxiv.org/pdf/1410.1465.pdf). 
This can be useful in some applications using inertial measurement units. Please see, [Associating Uncertainty to Extended Poses for on Lie Group IMU Preintegration with Rotating Earth](https://arxiv.org/abs/2007.14097), where the orientation, position and linear velocity (with the frame located in the origin of the body and the orientation same as the inertial frame) can be packed into a single Lie group.

Similarly. a possible extension could be towards a generic $SE_{N+2}(3)$ Lie group which can be useful for SLAM based applications. Please see [Contact-Aided Invariant Extended Kalman Filtering for Legged Robot State Estimation](https://arxiv.org/pdf/1805.10410.pdf).

@artivis Do you think it would be meaningful to have these Lie groups added to the repository? 

I've made an initial implementation for $SE_2(3)$ in my fork with https://github.com/prashanthr05/manif/commit/7169899e337ef0917377c83fd8e4d965704685fd and https://github.com/prashanthr05/manif/commit/c96f1f4465d10b30150e3f7fa3010f7a86dd2b41

However, there is a missing implementation for the action of this Lie group on a vector, `act()` method. It was not clear to me about how this operation might work. 

Let me know what you think about this. If you agree, I can maybe open a PR. Thank you!

artivis:
Hi @prashanthr05,

It would definitely be interesting to have `SE_2(3)` in `manif`! I've had my eyes on it for a time as a matter of fact. But I'm busy with a couple (upcoming) redesigns (starting with #150). However since you already implemented it, feel free to open a PR as it is, I will take care of 'porting' it to the new design once that's ready.

I'm not sure either about the `act` implementation, maybe @joansola has an idea? In case it does not make sense to implement it, we can put there a failing `static_assert` or such and handle it in the test suite.

While totally optional, it would be great to see a toy example making use of `SE_2(3)` accompanying the PR in the `examples` folder. That helps greatly in terms of documentation. Similarly don't hesitate to reference papers (and equations numbers) in the actual code documentation when possible.

To summarize, please do open a PR :+1:.

prashanthr05:
[...]

PR #147, opened 2020-07-25
Review copy construstor/assignment

artivis:
- Fix compilation warning in `lt::optional`
- Review all of the constructors and copy/move assignment (fixing a lot of warnings)
- Cleanup Eigen includes
- Add g++8/9 to CI

g++8/9 were issuing a lot of warnings concerning copy constructors and assignment due to implicit default being hidden in derived classes and such. This PR fixes that and (hopefully) brings some order in this matter. Some of the boilerplate code related to constructor/assignment is factored away in macros. 

Replaces #140 and #141 and #173. 

Closes #139
Closes #162
Closes #170
Closes #172

PR #154, opened 2020-08-03
Add SE_2(3) Lie group

prashanthr05:
This PR adds,
- $SE_2(3)$ Lie group (For more details, see #152)
- A $SE_2(3)$ localization with linear velocity estimation example, which is simply an extension of the $SE(3)$ localization example. 

P.S. I apologize for making it a single, huge PR.

The results obtained from a 10 step simulation of the $SE_2(3)$ localization example is shown below,
``` sh
X STATE     :    X      Y      Z    TH_x   TH_y   TH_z     V_x   V_y    V_z
---------------------------------------------------------------------------
X initial   : +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000
---------------------------------------------------------------------------
X simulated : +0.100 +0.000 +0.000 +0.000 +0.000 +0.050 +0.010 -0.000 +0.000
X estimated : +0.104 +0.009 -0.001 +0.005 -0.000 +0.046 +0.007 +0.005 -0.004
X unfilterd : +0.168 -0.021 +0.057 +0.060 +0.082 -0.010 +0.007 +0.005 -0.004
---------------------------------------------------------------------------
X simulated : +0.200 -0.000 +0.000 +0.000 +0.000 +0.100 +0.020 -0.000 +0.000
X estimated : +0.204 +0.007 -0.008 -0.001 -0.001 +0.096 +0.017 +0.012 -0.002
X unfilterd : +0.193 -0.041 -0.012 -0.018 +0.183 -0.011 +0.017 +0.012 -0.002
---------------------------------------------------------------------------
X simulated : +0.300 -0.000 +0.000 +0.000 +0.000 +0.150 +0.030 -0.000 +0.000
X estimated : +0.304 -0.002 +0.004 +0.000 -0.000 +0.149 +0.035 +0.010 -0.011
X unfilterd : +0.206 +0.041 -0.016 -0.094 +0.122 +0.080 +0.035 +0.008 -0.012
---------------------------------------------------------------------------
X simulated : +0.400 -0.000 +0.000 +0.000 +0.000 +0.200 +0.040 -0.000 +0.000
X estimated : +0.396 +0.002 -0.011 -0.001 -0.003 +0.200 +0.048 +0.007 -0.007
X unfilterd : +0.296 +0.124 +0.055 -0.042 +0.085 +0.177 +0.048 +0.004 -0.010
---------------------------------------------------------------------------
X simulated : +0.500 -0.000 +0.000 +0.000 +0.000 +0.250 +0.050 -0.000 +0.000
X estimated : +0.500 +0.005 +0.007 -0.001 +0.005 +0.248 +0.067 +0.016 -0.014
X unfilterd : +0.371 +0.198 +0.082 +0.040 +0.089 +0.166 +0.066 +0.014 -0.018
---------------------------------------------------------------------------
X simulated : +0.600 -0.000 +0.000 +0.000 +0.000 +0.300 +0.060 -0.000 +0.000
X estimated : +0.602 +0.001 +0.005 +0.001 +0.001 +0.301 +0.081 +0.013 -0.009
X unfilterd : +0.408 +0.102 +0.092 +0.090 +0.187 +0.303 +0.082 +0.008 -0.011
---------------------------------------------------------------------------
X simulated : +0.700 -0.000 +0.000 +0.000 +0.000 +0.350 +0.070 -0.000 +0.000
X estimated : +0.707 -0.000 -0.008 -0.001 -0.005 +0.351 +0.093 +0.019 -0.017
X unfilterd : +0.463 +0.026 +0.149 +0.146 +0.219 +0.443 +0.094 +0.009 -0.019
---------------------------------------------------------------------------
X simulated : +0.800 -0.000 +0.000 +0.000 +0.000 +0.400 +0.080 -0.000 +0.000
X estimated : +0.801 -0.018 -0.000 -0.002 +0.001 +0.405 +0.107 +0.012 -0.024
X unfilterd : +0.697 -0.058 +0.169 +0.098 +0.093 +0.577 +0.106 +0.000 -0.032
---------------------------------------------------------------------------
X simulated : +0.900 -0.000 +0.000 +0.000 +0.000 +0.450 +0.090 -0.000 +0.000
X estimated : +0.901 -0.005 -0.004 -0.001 -0.003 +0.450 +0.123 +0.020 -0.015
X unfilterd : +0.825 -0.029 +0.096 +0.134 +0.167 +0.713 +0.123 +0.004 -0.020
---------------------------------------------------------------------------
X simulated : +1.000 -0.000 +0.000 +0.000 +0.000 +0.500 +0.100 -0.000 +0.000
X estimated : +0.999 -0.002 -0.001 +0.001 -0.000 +0.501 +0.132 +0.028 -0.019
X unfilterd : +0.903 -0.082 -0.021 +0.187 +0.118 +0.786 +0.129 +0.011 -0.028
---------------------------------------------------------------------------

```

### DIFF
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -153,7 +153,7 @@ int main()
     // acceleration due to gravity in world frame
     Vector3d g;
     g << 0, 0, -9.80665;
-    double dt{0.1};
+    const double dt = 0.1;
 
     // IMU measurements in IMU frame
     Vector3d alpha, omega, alpha_prev, omega_prev;
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -232,11 +232,11 @@ int main()
 
     // IMU measurements in IMU frame
     Vector3d alpha, alpha_const, omega, alpha_prev, omega_prev;
-    alpha_const << 0.001, 0.02, 0.05; // constant acceleration in IMU frame without gravity compensation
-    omega << 0.02, 0.01, 0; // constant angular velocity about x- and y-direction in IMU frame
+    alpha_const << 0.1, 0.01, 0.1; // constant acceleration in IMU frame without gravity compensation
+    omega << 0.01, 0.1, 0; // constant angular velocity about x- and y-direction in IMU frame
 
     // Previous IMU measurements in IMU frame initialized to values expected when stationary
-    alpha_prev = alpha =  - (X_simulation.rotation()).transpose()*g;
+    alpha_prev = alpha = alpha_const - (X_simulation.rotation()).transpose()*g;
     omega_prev << 0, 0, 0;
 
     // Define a control vector and its noise and covariance
@@ -309,6 +309,7 @@ int main()
 
 
 
+
     // START TEMPORAL LOOP
     //
     //
@@ -323,9 +324,6 @@ int main()
         auto v_k = X_simulation.linearVelocity();
         auto acc_k = alpha_prev + R_k.transpose()*g;
 
-        /// update expected IMU measurements
-        alpha = alpha_const - R_k.transpose()*g;
-
         /// input vector
         u_nom << (dt*(R_k.transpose())*v_k + 0.5*dt*dt*acc_k),  dt*omega_prev, dt*acc_k;
 
@@ -338,6 +336,8 @@ int main()
 
         /// first we move - - - - - - - - - - - - - - - - - - - - - - - - - - - -
         X_simulation = X_simulation + u_simu;               // overloaded X.rplus(u) = X * exp(u)
+        /// update expected IMU measurements
+        alpha = alpha_const - X_simulation.rotation().transpose()*g; // update expected IMU measurement after moving
 
         /// then we measure all landmarks - - - - - - - - - - - - - - - - - - - -
         for (int i = 0; i < landmarks.size(); i++)
@@ -434,7 +434,9 @@ int main()
         cout << "---------------------------------------------------------------------------" << endl;
         // END DEBUG
 
+
     }
+
 
     //
     //
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -8,6 +8,9 @@
  *  This file is:
  *  (c) 2020 Prashanth Ramadoss @ DIC-IIT, Genova, Italy
  *
+ *  adapted from the file se3_localization.cpp:
+ *  (c) 2018 Joan Sola @ IRI-CSIC, Barcelona, Catalonia
+ *
  *  This file is part of `manif`, a C++ template-only library
  *  for Lie theory targeted at estimation for robotics.
  *  Manif is:
@@ -24,16 +27,16 @@
  *  We consider a robot in 3D space surrounded by a small
  *  number of punctual landmarks or _beacons_.
  *  The robot is assumed to be mounted with an IMU whose
- *  measurements are fed as exogeneous inputs to the system,
- *  and is able to measure the location
+ *  measurements are fed as exogeneous inputs to the system.
+ *  The robot is able to measure the location
  *  of the beacons w.r.t its own reference frame.
  *  We assume in this example that the IMU frame coincides with the robot frame.
  *
  *  The robot extended pose X is in SE_2(3) and the beacon positions b_k in R^3,
  *
- *      X = |  R   t  v|              // position, orientation and linear velocity
- *          |  0   1  0|
- *          |  0   0  1|
+ *      X = |    R   p  v|              // position, orientation and linear velocity
+ *          |0 0 0   1  0|
+ *          |0 0 0   0  1|
  *
  *      b_k = (bx_k, by_k, bz_k)    // lmk coordinates in world frame
  *
@@ -46,26 +49,73 @@
  * Note that the linear velocity v is expressed in a frame whose origin coincides with
  * the robot frame but has a orientation which is same as the world frame
  *
- *  The discrete-time system propagation can be written as,
- *   p_{k+1} = p_{k} + R_{k} (R_{k}.T v_{k} dt + 0.5*dt*dt*acc_{k})
- *   R_{k+1} = R_{k} Exp_SO3(omega_{k} dt + w_omega)
- *   v_{k+1} = v_{k} + R_{k} (acc_{k} dt + w_acc)
+ * Consider robot coordinate frame B and world coordinate frame A.
+ * - p is the position of the origin of the robot frame B with respect to the world frame A
+ * - R is the orientation of the robot frame B with respect to the world frame A
+ * - v is the velocity of the robot frame with respect to the world frame,
+ *            expressed in a frame whose origin coincides with the robot frame, oriented similar to the world frame
+ *            (it is equivalent to p_dot in continuous time. This is usually called mixed-frame representation
+ *            and is denoted as (B[A] v_AB), where B[A] is the mixed frame as described above.
+ *            For reference, please see "Multibody Dynamics Notation" by Silvio Traversaro and Alessandro Saccon.
+ *            Link: https://research.tue.nl/en/publications/multibody-dynamics-notation-version-2)
+ * - a is the frame acceleration in mixed-representation (equivalent to p_doubledot in continuous time).
+ * - omega_b as the angular velocity of the robot expressed in the robot frame
  *
- * Note that these equations are modified to obey the group composition rules.
+ * The kinematic equations (1) can be written as,
+ * p <-- p + v dt + 0.5 a dt^2
+ * R <-- R Exp_SO3(omega_b)
+ * v <-- v + a dt
  *
- * where,
- *        - R_{k}.T is the transpose of R_{k}
- *        - acc_{k} = alpha_{k} + R_{k}.T g
- *        - w_omega is the additive white noise affecting the gyroscope measurements
- *        - w_acc is the additive white noise affecting the linear accelerometer measurements
+ * We would like to express these equations in the form,
+ * X <-- X* Exp(u)
+ * where, X \in SE_2(3), u \in R^9 and u_hat \in se_2(3)
+ * Note that here input vector u is expressed in the local frame (robot frame).
  *
- *    Q = diagonal(0_{3x1}, 0_{3x1}, 0_{3x1}, sigma_omegax^2, sigma_omegay^2, sigma_omegaz^2, sigma_accx^2, sigma_accy^2, sigma_accz^2).
+ * The exponential mapping of SE_2(3) is defined as,
+ * for u = [u_p, u_w, u_v]
+ * Exp(u) = | Exp_SO3(u_w)   JlSO3(u_w) u_p   JlSO3(u_w) u_v |
+ *          | 0    0    0                 1                0 |
+ *          | 0    0    0                 0                1 |
+ * where, JlSO3 is the left Jacobian of the SO(3) group.
  *
- *  The exogenous input u becomes,
- *   u = (R_{k}.T v dt + 0.5*dt*dt*acc_{k}, omega_{k} dt, acc_{k} dt)
+ * Please see the Appendix C of the paper "A micro Lie theory for state estimation in robotics",
+ * for the definition of the left Jacobian of SO(3).
+ * Please see the Appendix D of the paper, for the definition of Exp map for SE(3).
+ * The Exp map of SE_2(3) is a simple extension from the Exp map of SE(3).
+ * Also, please refer to Example 7 of the paper to understand when and how the left Jacobian of SO(3)
+ * appears in the definitions of Exp maps. The Example 7 illustrates the scenario for SE(3).
+ * We use a direct extension here for SE_2(3).
+ * One can arrive to such a definition by following the convergent Taylor's series expansion
+ * for the matrix exponential of the Lie algebra element (Equation 16 of the paper).
  *
- * Note that all the elements of the input vector u belong in the IMU frame/robot frame.
+ * As a result of X <-- X * Exp(u), we get (2)
+ * p <-- p + R JlSO3(u_w) u_p
+ * R <-- R Exp_SO3(u_w)
+ * v <-- v + R JlSO3(u_w) u_v
  *
+ * We would like to draw a relationship between the sets of equations (1) and (2).
+ * Additionally, we would also like to make use of the IMU measurements as exogeneous inputs.
+ * Considering R.T as the transpose of R, the IMU measurements are modeled as,
+ *    - linear accelerometer measurements alpha = R.T (a - g) + w_acc
+ *    - gyroscope measurements omega = omega_b + w_omega
+ * Note that the IMU measurements are expressed in the IMU frame (coincides with the robot frame - assumption).
+ * The IMU measurements are corrupted by noise,
+ *    - w_omega is the additive white noise affecting the gyroscope measurements
+ *    - w_acc is the additive white noise affecting the linear accelerometer measurements
+ *
+ * Taking into account all of the above considerations, the exogenous input vector u (3) becomes,
+ *   u = (u_p, u_w, u_v) where,
+ *   u_w = omega dt
+ *   Jinv = inv(JlSO3(u_w)) is the left Jacobian inverse of SO(3) Lie group
+ *   u_p = Jinv * (R.T v dt + 0.5 dt^2 (alpha + R.T g))
+ *   u_v = Jinv * ((alpha + R.T g) dt)
+ *
+ * To cross verify the equations, substituting (3) in (2) will give us (1).
+ * This choice of input vector allows us to directly use measurements from the IMU
+ * for an unified motion integration involving position, orientation and linear velocity of the robot.
+ *
+ * The system propagation noise covariance amtrix becomes,
+ *    U = diagonal(0, 0, 0, sigma_omegax^2, sigma_omegay^2, sigma_omegaz^2, sigma_accx^2, sigma_accy^2, sigma_accz^2).
  *
  *  At the arrival of a exogeneous input u, the robot pose is updated
  *  with X <-- X * Exp(u) = X + u.
@@ -86,15 +136,15 @@
  *  All these variables are summarized again as follows
  *
  *    X   : robot's extended pose, SE_2(3)
- *    u   : robot control input, (R_{k}.T v dt + 0.5*dt*dt*acc_{k}, omega_{k} dt, acc_{k} dt)
- *    Q   : control perturbation covariance
+ *    u   : robot control input, (Jinv * (R.T v dt + 0.5 dt^2 (alpha + R.T g)), omega dt, Jinv * ((alpha + R.T g) dt))
+ *    U   : control perturbation covariance
  *    b_k : k-th landmark position, R^3
  *    y   : Cartesian landmark measurement in robot frame, R^3
  *    R   : covariance of the measurement noise
  *
  *  The motion and measurement models are
  *
- *    X_(t+1) = f(X_t, u) = X_t * Exp ( w )     // motion equation
+ *    X_(t+1) = f(X_t, u) = X_t * Exp ( u )     // motion equation
  *    [y_k; 1; 0]     = h(X, b_k) = X^-1 * [b_k; 1; 0]          // measurement equation
  *
  *  The algorithm below comprises first a simulator to
@@ -107,6 +157,13 @@
  *  Printing simulated state and estimated state together
  *  with an unfiltered state (i.e. without Kalman corrections)
  *  allows for evaluating the quality of the estimates.
+ *
+ * A side note: Besides the approach described here in this illustration example,
+ * there are other interesting works like the paper,
+ * The Invariant Extended Kalman filter as a stable observer (https://arxiv.org/pdf/1410.1465.pdf)
+ * which assume a specific structure for the system propagation dynamics "f(X_t, u)" (group affine dynamics)
+ * that simplifies the covariance propagation and enables error dynamics with stronger convergence properties.
+ *
  */
 
 #include "manif/SE_2_3.h"
@@ -157,8 +214,9 @@ int main()
 
     // IMU measurements in IMU frame
     Vector3d alpha, omega, alpha_prev, omega_prev;
-    alpha << 10.0, 0.0, 9.80665;
-    omega << 1, 0.1, 0;
+    alpha << 10.0, 0.0, 9.80665; // constant acceleration along x-direction
+    omega << 1, 0.1, 0; // constant angular velocity about x- and y-direction
+    // Previous IMU measurements in IMU frame initialized to values expected when stationary
     alpha_prev << 0.0, 0.0, 9.80665;
     omega_prev << 0, 0, 0;
 
@@ -234,13 +292,15 @@ int main()
     {
         //// I. Simulation ###############################################################################
 
-        /// get current state
+        /// get current state and measurements from previous step
         auto R_k = X_simulation.rotation();
         auto v_k = X_simulation.linearVelocity();
         auto acc_k = alpha_prev + R_k.transpose()*g;
+        manif::SO3Tangentd Rtan = dt*omega_prev;
+        auto Jinv = Rtan.ljacinv();
 
         /// state dependent velocity vector
-        u_nom << (dt*(R.transpose())*v_k + 0.5*dt*dt*acc_k),  dt*omega_prev, dt*acc_k;
+        u_nom << Jinv*(dt*(R.transpose())*v_k + 0.5*dt*dt*acc_k),  dt*omega_prev, Jinv*dt*acc_k;
 
         /// simulate noise
         u_noise = u_sigmas * Array9d::Random();             // control noise
@@ -263,7 +323,7 @@ int main()
 
             y = X_simulation.inverse().act(b);              // landmark measurement, before adding noise
             y = y + y_noise;                                // landmark measurement, noisy
-            measurements[i] << y, 1, 0;                            // store for the estimator just below
+            measurements[i] << y, 1, 0;                     // store for the estimator just below
         }
 
 
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -137,7 +137,7 @@
  *
  *  We consider the beacons b_k situated at known positions.
  *  We define the extended pose to estimate as X in SE_2(3).
- *  The estimation error (innovation) dx and its covariance P are expressed
+ *  The estimation error dx and its covariance P are expressed
  *  in the tangent space at X.
  *
  *  All these variables are summarized again as follows
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -35,8 +35,8 @@
  *  The robot extended pose X is in SE_2(3) and the beacon positions b_k in R^3,
  *
  *      X = |    R   p  v|              // position, orientation and linear velocity
- *          |0 0 0   1  0|
- *          |0 0 0   0  1|
+ *          |        1   |
+ *          |           1|
  *
  *      b_k = (bx_k, by_k, bz_k)    // lmk coordinates in world frame
  *
@@ -67,7 +67,7 @@
  * v <-- v + a dt
  *
  * We would like to express these equations in the form,
- * X <-- X* Exp(u)
+ * X <-- X * Exp(u)
  * where, X \in SE_2(3), u \in R^9 and u_hat \in se_2(3)
  * Note that here input vector u is expressed in the local frame (robot frame).
  *
@@ -102,6 +102,7 @@
  * The IMU measurements are corrupted by noise,
  *    - w_omega is the additive white noise affecting the gyroscope measurements
  *    - w_acc is the additive white noise affecting the linear accelerometer measurements
+ * It must be noted that we do not consider IMU biases in the IMU measurement model in this example.
  *
  * Taking into account all of the above considerations, the exogenous input vector u (3) becomes,
  *   u = (u_p, u_w, u_v) where,
@@ -114,7 +115,7 @@
  * This choice of input vector allows us to directly use measurements from the IMU
  * for an unified motion integration involving position, orientation and linear velocity of the robot.
  *
- * The system propagation noise covariance amtrix becomes,
+ * The system propagation noise covariance matrix becomes,
  *    U = diagonal(0, 0, 0, sigma_omegax^2, sigma_omegay^2, sigma_omegaz^2, sigma_accx^2, sigma_accy^2, sigma_accz^2).
  *
  *  At the arrival of a exogeneous input u, the robot pose is updated
@@ -124,19 +125,25 @@
  *  though they are put in Cartesian form for simplicity.
  *  Their noise n is zero mean Gaussian, and is specified
  *  with a covariances matrix R.
- *  We notice the rigid motion action [y 0 0]' = h(X,b) = X^-1 * [b 1 0]'
+ *  We notice that the SE_2(3) action is the same as a
+ *  rigid motion action of SE(3).
+ *  This is the action of X \in SE_2(3) on a 3-d point b \in R^3 defined as,
+ *  X b = R b + p
+ *
+ *  Thus, the landmark measurements can be expressed as,
+ *  y = h(X,b) = X^-1 * b'
  *
  *      y_k = (brx_k, bry_k, brz_k)    // lmk coordinates in robot frame
  *
  *  We consider the beacons b_k situated at known positions.
  *  We define the extended pose to estimate as X in SE_2(3).
- *  The estimation error dx and its covariance P are expressed
+ *  The estimation error (innovation) dx and its covariance P are expressed
  *  in the tangent space at X.
  *
  *  All these variables are summarized again as follows
  *
  *    X   : robot's extended pose, SE_2(3)
- *    u   : robot control input, (Jinv * (R.T v dt + 0.5 dt^2 (alpha + R.T g)), omega dt, Jinv * ((alpha + R.T g) dt))
+ *    u   : robot control input, u = u(X, y_imu) \in se_2(3) with X as state and y_imu = [alpha, omega] as IMU readings, see Eq. (3)
  *    U   : control perturbation covariance
  *    b_k : k-th landmark position, R^3
  *    y   : Cartesian landmark measurement in robot frame, R^3
@@ -145,7 +152,7 @@
  *  The motion and measurement models are
  *
  *    X_(t+1) = f(X_t, u) = X_t * Exp ( u )     // motion equation
- *    [y_k; 1; 0]     = h(X, b_k) = X^-1 * [b_k; 1; 0]          // measurement equation
+ *    y_k     = h(X, b_k) = X^-1 * b_k          // measurement equation
  *
  *  The algorithm below comprises first a simulator to
  *  produce measurements, then uses these measurements
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -304,7 +304,7 @@ int main()
     //
 
     // Make 10 steps. Measure up to three landmarks each time.
-    for (int t = 0; t < 10; t++)
+    for (double t = 0; t < 10; t+=dt)
     {
         //// I. Simulation ###############################################################################
 
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -188,7 +188,6 @@
 #include <Eigen/Dense>
 
 #include <vector>
-
 #include <iostream>
 #include <iomanip>
 
@@ -205,6 +204,15 @@ typedef Array<double, 9, 1> Array9d;
 typedef Matrix<double, 9, 1> Vector9d;
 typedef Matrix<double, 9, 9> Matrix9d;
 
+
+Eigen::Matrix3d skew(Eigen::Ref<const Eigen::Vector3d> t)
+{
+    Eigen::Matrix3d tcross;
+    tcross << 0, -t(2),  t(1),
+           t(2),     0, -t(0),
+          -t(1),  t(0),    0;
+    return tcross;
+}
 
 int main()
 {
@@ -228,7 +236,7 @@ int main()
     // acceleration due to gravity in world frame
     Vector3d g;
     g << 0, 0, -9.80665;
-    const double dt = 0.1;
+    const double dt = 0.01;
 
     // IMU measurements in IMU frame
     Vector3d alpha, alpha_const, omega, alpha_prev, omega_prev;
@@ -249,7 +257,7 @@ int main()
     U        = (u_sigmas * u_sigmas).matrix().asDiagonal();
 
     // Declare the Jacobians of the motion wrt robot and control
-    manif::SE_2_3d::Jacobian J_x, J_u;
+    manif::SE_2_3d::Jacobian slantFk, Fk, J_x, J_u;
 
     // Define five landmarks in R^3
     Vector3d b0, b1, b2, b3, b4, b;
@@ -308,8 +316,6 @@ int main()
     // END DEBUG
 
 
-
-
     // START TEMPORAL LOOP
     //
     //
@@ -362,14 +368,26 @@ int main()
         auto v_k_est = X.linearVelocity();
         auto acc_k_est = alpha_prev + R_k_est.transpose()*g;
 
-        u_est << (dt*(R_k_est.transpose())*v_k_est + 0.5*dt*dt*acc_k_est),  dt*omega_prev, dt*acc_k_est;
+        Eigen::Vector3d accLin = dt*(R_k_est.transpose())*v_k_est + 0.5*dt*dt*acc_k_est;
+        auto gLin = R_k_est.transpose()*g*dt;
+        Eigen::Matrix3d accLinCross = skew(accLin);
+        Eigen::Matrix3d gCross = skew(gLin);
+
+        u_est << accLin,  dt*omega_prev, dt*acc_k_est;
         u_est += u_noise;
 
         /// First we move - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
         X = X.plus(u_est, J_x, J_u);                        // X * exp(u), with Jacobians
 
-        P = J_x * P * J_x.transpose() + J_u * U * J_u.transpose();
+        // Prepare Jacobian
+        slantFk.setZero();
+        slantFk.block<3, 3>(0, 3) = accLinCross;
+        slantFk.block<3, 3>(0, 6) = Eigen::Matrix3d::Identity()*dt;
+        slantFk.block<3, 3>(6, 3) = gCross;
+        Fk = J_x + J_u*slantFk;
+
+        P = Fk * P * Fk.transpose() + J_u * U * J_u.transpose();
 
 
         /// Then we correct using the measurements of each lmk - - - - - - - - -
@@ -434,9 +452,7 @@ int main()
         cout << "---------------------------------------------------------------------------" << endl;
         // END DEBUG
 
-
     }
-
 
     //
     //
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -247,7 +247,10 @@ int main()
     U        = (u_sigmas * u_sigmas).matrix().asDiagonal();
 
     // Declare the Jacobians of the motion wrt robot and control
-    manif::SE_2_3d::Jacobian slantFk, Fk, J_x, J_u;
+    manif::SE_2_3d::Jacobian F;     // F = J_x_x + (J_x_u * J_u_x)
+    manif::SE_2_3d::Jacobian J_x_x; // d(X * exp(u)) / dX
+    manif::SE_2_3d::Jacobian J_x_u; // d(X * exp(u)) / du
+    manif::SE_2_3d::Jacobian J_u_x; // du / dX, since u is a state-dependent vector
 
     // Define five landmarks in R^3
     Vector3d b0, b1, b2, b3, b4, b;
@@ -368,16 +371,16 @@ int main()
 
         /// First we move - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
-        X = X.plus(u_est, J_x, J_u);                        // X * exp(u), with Jacobians
+        X = X.plus(u_est, J_x_x, J_x_u);                        // X * exp(u), with Jacobians
 
-        // Prepare Jacobian
-        slantFk.setZero();
-        slantFk.block<3, 3>(0, 3) = accLinCross;
-        slantFk.block<3, 3>(0, 6) = Eigen::Matrix3d::Identity()*dt;
-        slantFk.block<3, 3>(6, 3) = gCross;
-        Fk = J_x + J_u*slantFk;
+        // Prepare Jacobian of state-dependent control vector
+        J_u_x.setZero();
+        J_u_x.block<3, 3>(0, 3) = accLinCross;
+        J_u_x.block<3, 3>(0, 6) = Eigen::Matrix3d::Identity()*dt;
+        J_u_x.block<3, 3>(6, 3) = gCross;
+        F = J_x_x + J_x_u*J_u_x;                                // chain rule for system model Jacobian
 
-        P = Fk * P * Fk.transpose() + J_u * U * J_u.transpose();
+        P = F * P * F.transpose() + J_x_u * U * J_x_u.transpose();
 
 
         /// Then we correct using the measurements of each lmk - - - - - - - - -
--- a/examples/se_2_3_localization.cpp
+++ b/examples/se_2_3_localization.cpp
@@ -1,0 +1,302 @@
+/**
+ * \file se_2_3_localization.cpp
+ *
+ *  Created on: Aug 03, 2020
+ *     \author: prashanthr05
+ *
+ *  ---------------------------------------------------------
+ *  This file is:
+ *  (c) 2018 Joan Sola @ IRI-CSIC, Barcelona, Catalonia
+ *
+ *  This file is part of `manif`, a C++ template-only library
+ *  for Lie theory targeted at estimation for robotics.
+ *  Manif is:
+ *  (c) 2018 Jeremie Deray @ IRI-UPC, Barcelona
+ *  ---------------------------------------------------------
+ *
+ *  ---------------------------------------------------------
+ *  Demonstration example:
+ *
+ *  3D Robot localization and linear velocity estimation based on fixed beacons.
+ *
+ *  ---------------------------------------------------------
+ *
+ *  We consider a robot in 3D space surrounded by a small
+ *  number of punctual landmarks or _beacons_.
+ *  The robot receives control actions in the form of axial
+ *  and angular velocities, and is able to measure the location
+ *  of the beacons w.r.t its own reference frame.
+ *
+ *  The robot extended pose X is in SE_2(3) and the beacon positions b_k in R^3,
+ *
+ *      X = |  R   t  v|              // position, orientation and linear velocity
+ *          |  0   1  0|
+ *          |  0   0  1|
+ *
+ *      b_k = (bx_k, by_k, bz_k)    // lmk coordinates in world frame
+ *
+ * Note that the linear velocity is expressed in a frame whose origin coincides with
+ * the robot frame but has a orientation which is same as the world frame
+ *
+ *  The control signal u is a vector in se_2(3) comprising longitudinal
+ *  velocity vx and angular velocity wz, with no other velocity
+ *  components or acceleration components, integrated over the sampling time dt.
+ *
+ *      u = (vx*dt, 0, 0, 0, 0, w*dt, 0, 0, 0)
+ *
+ *  The control is corrupted by additive Gaussian noise u_noise,
+ *  with covariance
+ *
+ *    Q = diagonal(sigma_x^2, sigma_y^2, sigma_z^2, sigma_roll^2, sigma_pitch^2, sigma_yaw^2, sigma_vx^2, sigma_vy^2, sigma_vz^2).
+ *
+ *  This noise accounts for possible lateral and rotational slippage
+ *  through a non-zero values of sigma_y, sigma_z, sigma_roll and sigma_pitch.
+ *
+ *  At the arrival of a control u, the robot pose is updated
+ *  with X <-- X * Exp(u) = X + u.
+ *
+ *  Landmark measurements are of the range and bearing type,
+ *  though they are put in Cartesian form for simplicity.
+ *  Their noise n is zero mean Gaussian, and is specified
+ *  with a covariances matrix R.
+ *  We notice the rigid motion action [y 0 0]' = h(X,b) = X^-1 * [b 1 0]'
+ *
+ *      y_k = (brx_k, bry_k, brz_k)    // lmk coordinates in robot frame
+ *
+ *  We consider the beacons b_k situated at known positions.
+ *  We define the extended pose to estimate as X in SE_2(3).
+ *  The estimation error dx and its covariance P are expressed
+ *  in the tangent space at X.
+ *
+ *  All these variables are summarized again as follows
+ *
+ *    X   : robot's extended pose, SE_2(3)
+ *    u   : robot control, (v*dt; 0; 0; 0; 0; w*dt; 0; 0; 0) in se_2(3)
+ *    Q   : control perturbation covariance
+ *    b_k : k-th landmark position, R^3
+ *    y   : Cartesian landmark measurement in robot frame, R^3
+ *    R   : covariance of the measurement noise
+ *
+ *  The motion and measurement models are
+ *
+ *    X_(t+1) = f(X_t, u) = X_t * Exp ( w )     // motion equation
+ *    [y_k; 1; 0]     = h(X, b_k) = X^-1 * [b_k; 1; 0]          // measurement equation
+ *
+ *  The algorithm below comprises first a simulator to
+ *  produce measurements, then uses these measurements
+ *  to estimate the state, using a Lie-based error-state Kalman filter.
+ *
+ *  This file has plain code with only one main() function.
+ *  There are no function calls other than those involving `manif`.
+ *
+ *  Printing simulated state and estimated state together
+ *  with an unfiltered state (i.e. without Kalman corrections)
+ *  allows for evaluating the quality of the estimates.
+ */
+
+#include "manif/SE_2_3.h"
+
+#include <Eigen/Dense>
+
+#include <vector>
+
+#include <iostream>
+#include <iomanip>
+
+using std::cout;
+using std::endl;
+
+using namespace Eigen;
+
+typedef Array<double, 3, 1> Array3d;
+typedef Matrix<double, 5, 1> Vector5d;
+typedef Array<double, 6, 1> Array6d;
+typedef Matrix<double, 6, 1> Vector6d;
+typedef Matrix<double, 6, 6> Matrix6d;
+typedef Array<double, 9, 1> Array9d;
+typedef Matrix<double, 9, 1> Vector9d;
+typedef Matrix<double, 9, 9> Matrix9d;
+
+
+int main()
+{
+    // START CONFIGURATION
+    //
+    //
+    const int NUMBER_OF_LMKS_TO_MEASURE = 5;
+
+    // Define the robot extended pose element and its covariance
+    manif::SE_2_3d X, X_simulation, X_unfiltered;
+    Matrix9d    P;
+
+    X_simulation.setIdentity();
+    X.setIdentity();
+    X_unfiltered.setIdentity();
+    P.setZero();
+
+    // Define a control vector and its noise and covariance
+    manif::SE_2_3Tangentd  u_simu, u_est, u_unfilt;
+    Vector9d            u_nom, u_noisy, u_noise;
+    Array9d             u_sigmas;
+    Matrix9d            U;
+
+    u_nom    << 0.1, 0.0, 0.0, 0.0, 0.0, 0.05, 0.01, 0.0, 0.0;
+    u_sigmas << 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.01, 0.01, 0.01;
+    U        = (u_sigmas * u_sigmas).matrix().asDiagonal();
+
+    // Declare the Jacobians of the motion wrt robot and control
+    manif::SE_2_3d::Jacobian J_x, J_u;
+
+    // Define five landmarks in R^3
+    Vector3d b0, b1, b2, b3, b4, b;
+    b0 << 2.0,  0.0,  0.0;
+    b1 << 3.0, -1.0, -1.0;
+    b2 << 2.0, -1.0,  1.0;
+    b3 << 2.0,  1.0,  1.0;
+    b4 << 2.0,  1.0, -1.0;
+    std::vector<Vector3d> landmarks;
+    landmarks.push_back(b0);
+    landmarks.push_back(b1);
+    landmarks.push_back(b2);
+    landmarks.push_back(b3);
+    landmarks.push_back(b4);
+
+    // Define the beacon's measurements
+    Vector3d                y, y_noise;
+    Array3d                 y_sigmas;
+    Matrix3d                R;
+    std::vector<Vector5d>   measurements(landmarks.size());
+
+    y_sigmas << 0.01, 0.01, 0.01;
+    R        = (y_sigmas * y_sigmas).matrix().asDiagonal();
+
+    // Declare the Jacobian of the measurements wrt the robot pose
+    Matrix<double, 3, 9>    H;      // H = J_e_x
+
+    // Declare some temporaries
+    Vector3d                e, z;   // expectation, innovation
+    Matrix3d                E, Z;   // covariances of the above
+    Matrix<double, 9, 3>    K;      // Kalman gain
+    manif::SE_2_3Tangentd      dx;     // optimal update step, or error-state
+    manif::SE_2_3d::Jacobian   J_xi_x; // Jacobian is typedef Matrix
+    Matrix<double, 3, 9>       J_e_xi; // Jacobian
+
+    //
+    //
+    // CONFIGURATION DONE
+
+
+
+    // DEBUG
+    cout << std::fixed   << std::setprecision(3) << std::showpos << endl;
+    cout << "X STATE     :    X      Y      Z    TH_x   TH_y   TH_z     V_x   V_y    V_z" << endl;
+    cout << "---------------------------------------------------------------------------" << endl;
+    cout << "X initial   : " << X_simulation.log().coeffs().transpose() << endl;
+    cout << "---------------------------------------------------------------------------" << endl;
+    // END DEBUG
+
+
+
+
+    // START TEMPORAL LOOP
+    //
+    //
+
+    // Make 10 steps. Measure up to three landmarks each time.
+    for (int t = 0; t < 10; t++)
+    {
+        //// I. Simulation ###############################################################################
+
+        /// simulate noise
+        u_noise = u_sigmas * Array9d::Random();             // control noise
+        u_noisy = u_nom + u_noise;                          // noisy control
+
+        u_simu   = u_nom;
+        u_est    = u_noisy;
+        u_unfilt = u_noisy;
+
+        /// first we move - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+        X_simulation = X_simulation + u_simu;               // overloaded X.rplus(u) = X * exp(u)
+
+        /// then we measure all landmarks - - - - - - - - - - - - - - - - - - - -
+        for (int i = 0; i < landmarks.size(); i++)
+        {
+            b = landmarks[i];                               // lmk coordinates in world frame
+
+            /// simulate noise
+            y_noise = y_sigmas * Array3d::Random();         // measurement noise
+
+            y = X_simulation.inverse().act(b);              // landmark measurement, before adding noise
+            y = y + y_noise;                                // landmark measurement, noisy
+            measurements[i] << y, 1, 0;                            // store for the estimator just below
+        }
+
+
+
+
+        //// II. Estimation ###############################################################################
+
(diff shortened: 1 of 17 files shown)


examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,302 @@
...
+ *
+ *  3D Robot localization and linear velocity estimation based on fixed beacons.
+ *
+ *  ---------------------------------------------------------
+ *
+ *  We consider a robot in 3D space surrounded by a small
+ *  number of punctual landmarks or _beacons_.
+ *  The robot receives control actions in the form of axial
+ *  and angular velocities, and is able to measure the location
+ *  of the beacons w.r.t its own reference frame.
```
joansola, line 28: This doc should be updated

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,302 @@
...
+ *
+ *  ---------------------------------------------------------
+ *  This file is:
+ *  (c) 2018 Joan Sola @ IRI-CSIC, Barcelona, Catalonia
+ *
+ *  This file is part of `manif`, a C++ template-only library
+ *  for Lie theory targeted at estimation for robotics.
+ *  Manif is:
+ *  (c) 2018 Jeremie Deray @ IRI-UPC, Barcelona
+ *  ---------------------------------------------------------
```
joansola, line 15: This authorship should be updated

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,302 @@
...
+ *
+ * Note that the linear velocity is expressed in a frame whose origin coincides with
+ * the robot frame but has a orientation which is same as the world frame
+ *
+ *  The control signal u is a vector in se_2(3) comprising longitudinal
+ *  velocity vx and angular velocity wz, with no other velocity
+ *  components or acceleration components, integrated over the sampling time dt.
+ *
+ *      u = (vx*dt, 0, 0, 0, 0, w*dt, 0, 0, 0)
+ *
```
joansola, line 46: I guess control `u` should be IMU measurements

include/manif/impl/se_2_3/SE_2_3Tangent_base.h
```diff
@@ -0,0 +1,394 @@
...
+
+};
+
+template <typename _Derived>
+typename SE_2_3TangentBase<_Derived>::LieGroup
+SE_2_3TangentBase<_Derived>::exp(OptJacobianRef J_m_t) const
+{
+  using std::sqrt;
+  using std::cos;
+  using std::sin;
```
artivis, line 120: It looks like you don't need those, please remove them.

examples/se_2_3_localization.cpp
prashanthr05: I was actually wondering about it; whether to pass the IMU measurements as exogenous inputs or not.
I rather chose this simpler case instead. 

I will modify this to use expected IMU measurements.

include/manif/impl/se_2_3/SE_2_3_base.h
```diff
@@ -0,0 +1,448 @@
...
+  /// this is
+  ///       Ad(g) = | R T.R 0|
+  ///               | 0  R  0|
+  ///               | 0 V.R R|
+  ///
+  /// considering vee(log(g)) = (v;w;a)
+  /// with T = [t]_x
+  /// with V = [v]_x
+
+  Jacobian Adj = Jacobian::Zero();
```
artivis, line 344: Not important but this sets the entire matrix to zero and then rewrite half of it. I wonder if setting only the other half to zero would be an improvement.

examples/se_2_3_localization.cpp
joansola: I do not know the mechanisms of this SE23 group. We had invented a different Lie group for IMU, see here https://upcommons.upc.edu/handle/2117/187253, where the time interval is part of the group too.

In our case, integrating motion is done as follows:

1. `u = [w*dt, 0, a*dt, dt] \in R^10`
2. `X(t+dt) = X(t) * Exp(u)`

In your case I do not know how you handle time so I cannot suggest an implementation for this part of the code commented here. 

Nevertheless the idea of these examples is that they be realistic, and I think that the inputs to this example should be the IMU measurements `a` and `w`, together with the time interval `dt`.

include/manif/impl/se3/SE3Tangent_base.h
```diff
@@ -125,7 +125,7 @@ struct SE3TangentBase : TangentBase<_Derived>
     return Eigen::Map<SO3Tangent<Scalar>>(coeffs.data()+3);
   }
 
-private:
+// private:
```
prashanthr05, line 128: @artivis  Is it ok that I commented out the `private` specifier here? Do you recommend that I remove it?

examples/se_2_3_localization.cpp
prashanthr05: Improved the example in https://github.com/artivis/manif/pull/154/commits/945c4728639519d7eaa7bb47e7de7eacb9171fab.

using discrete-time propagation compatible with SE_2(3) composition,
```
p_{k+1} = p_{k} + R_{k} (R_{k}.T v_{k} dt + 0.5*dt*dt*acc_{k})
R_{k+1} = R_{k} Exp_SO3(omega_{k} dt + w_omega)
v_{k+1} = v_{k} + R_{k} (acc_{k} dt + w_acc)
```

The input vector `u` becomes state-dependent. 
The measurements remain the same beacon measurements.

Please let me know what you think about it.

For IMU measurements, 
```
alpha << 10.0, 0.0, 9.80665;
omega << 1, 0.1, 0;
g << 0, 0, -9.80665;
```

The results are as follows, 

``` sh
X STATE     :    X      Y      Z    TH_x   TH_y   TH_z     V_x   V_y    V_z
---------------------------------------------------------------------------
X initial   : +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000
---------------------------------------------------------------------------
X simulated : +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000
X estimated : +0.004 +0.009 -0.002 +0.005 -0.000 -0.005 -0.033 +0.054 -0.044
X unfilterd : +0.000 +0.000 +0.000 +0.060 +0.082 -0.060 -0.033 +0.054 -0.044
---------------------------------------------------------------------------
X simulated : +0.050 +0.000 +0.000 +0.100 +0.010 +0.000 +1.000 +0.000 +0.000
X estimated : +0.054 +0.008 -0.005 +0.099 +0.009 -0.005 +0.969 +0.116 -0.023
X unfilterd : +0.050 -0.001 -0.002 +0.082 +0.193 -0.114 +0.974 +0.090 -0.064
---------------------------------------------------------------------------
X simulated : +0.101 -0.005 -0.000 +0.200 +0.020 +0.000 +2.010 -0.098 +0.000
X estimated : +0.104 -0.001 -0.002 +0.201 +0.018 -0.004 +2.056 -0.010 -0.120
X unfilterd : +0.100 -0.010 -0.008 +0.100 +0.139 -0.086 +2.039 -0.112 -0.282
---------------------------------------------------------------------------
X simulated : +0.152 -0.015 +0.000 +0.300 +0.030 +0.000 +3.029 -0.295 +0.005
X estimated : +0.153 -0.012 -0.004 +0.299 +0.029 -0.000 +3.104 -0.250 -0.081
X unfilterd : +0.150 -0.025 -0.012 +0.245 +0.097 -0.050 +3.067 -0.452 -0.336
---------------------------------------------------------------------------
X simulated : +0.203 -0.029 +0.001 +0.400 +0.040 -0.000 +4.059 -0.590 +0.020
X estimated : +0.204 -0.026 -0.001 +0.401 +0.040 -0.001 +4.222 -0.454 -0.129
X unfilterd : +0.202 -0.038 -0.012 +0.425 +0.106 -0.114 +4.193 -0.622 -0.390
---------------------------------------------------------------------------
X simulated : +0.255 -0.049 +0.002 +0.500 +0.050 -0.000 +5.098 -0.984 +0.050
X estimated : +0.256 -0.046 +0.001 +0.502 +0.049 -0.001 +5.299 -0.889 -0.065
X unfilterd : +0.251 -0.071 -0.002 +0.587 +0.185 -0.014 +5.233 -1.346 -0.160
---------------------------------------------------------------------------
X simulated : +0.308 -0.074 +0.005 +0.600 +0.060 -0.000 +6.148 -1.477 +0.100
X estimated : +0.310 -0.070 +0.002 +0.598 +0.059 -0.002 +6.373 -1.314 -0.085
X unfilterd : +0.300 -0.109 +0.001 +0.742 +0.198 +0.076 +6.243 -2.041 -0.174
---------------------------------------------------------------------------
X simulated : +0.361 -0.104 +0.009 +0.700 +0.070 -0.000 +7.207 -2.070 +0.175
X estimated : +0.362 -0.102 +0.007 +0.701 +0.070 -0.003 +7.451 -1.946 -0.099
X unfilterd : +0.350 -0.145 -0.022 +0.766 +0.082 +0.121 +7.235 -2.821 -0.724
---------------------------------------------------------------------------
X simulated : +0.414 -0.138 +0.014 +0.800 +0.080 -0.000 +8.276 -2.762 +0.281
X estimated : +0.415 -0.137 +0.011 +0.800 +0.078 -0.002 +8.597 -2.625 +0.110
X unfilterd : +0.399 -0.197 -0.003 +0.915 +0.132 +0.228 +8.317 -3.862 -0.234
---------------------------------------------------------------------------
X simulated : +0.468 -0.178 +0.021 +0.900 +0.090 -0.000 +9.355 -3.555 +0.424
X estimated : +0.469 -0.178 +0.019 +0.900 +0.090 +0.000 +9.675 -3.328 +0.249
X unfilterd : +0.456 -0.236 -0.005 +1.048 +0.092 +0.228 +9.425 -4.529 -0.293
---------------------------------------------------------------------------

```

include/manif/impl/se_2_3/SE_2_3Tangent_base.h
prashanthr05: Fixed in https://github.com/artivis/manif/pull/154/commits/810862f30d3e97e9de9dbd54e1a7458146ae91f5

include/manif/impl/se_2_3/SE_2_3_base.h
prashanthr05: Fixed in https://github.com/artivis/manif/pull/154/commits/810862f30d3e97e9de9dbd54e1a7458146ae91f5

examples/se_2_3_localization.cpp
prashanthr05: Fixed in https://github.com/artivis/manif/pull/154/commits/945c4728639519d7eaa7bb47e7de7eacb9171fab

examples/se_2_3_localization.cpp
prashanthr05: Fixed in https://github.com/artivis/manif/pull/154/commits/945c4728639519d7eaa7bb47e7de7eacb9171fab

include/manif/impl/se3/SE3Tangent_base.h
artivis: You can leave it as it is, I am planning on [moving this function outside the class in the future](https://github.com/artivis/manif/blob/d33e100dd0acd9d6f48ecff69169114481aa8656/include/manif/impl/se3/SE3_utils.h#L8).

artivis:
@prashanthr05 Could you also update the `README.md` to add references to both the new group and accompanying example? See [this section](https://github.com/artivis/manif/blob/1d49ccc1937b2e204c3c559dad270110689f17ff/README.md#package-summary), right above the [doc section](https://github.com/artivis/manif/blob/1d49ccc1937b2e204c3c559dad270110689f17ff/README.md#documentation) and the [tuto section](https://github.com/artivis/manif/blob/1d49ccc1937b2e204c3c559dad270110689f17ff/README.md#tutorials-and-application-demos).
Furthermore, only if you wish so, you could add a reference to the `lie-group-controllers` in the `projects.md` file (consider a 1-2 lines description and a link).

prashanthr05:
> @prashanthr05 Could you also update the README.md to add references to both the new group and accompanying example? See this section, right above the doc section and the tuto section.

Sure.

> Furthermore, only if you wish so, you could add a reference to the lie-group-controllers in the projects.md file (consider a 1-2 lines description and a link).

Regarding this, we might have to ask @GiulioRomualdi :)

prashanthr05:
>>@prashanthr05 Could you also update the README.md to add references to both the new group and accompanying example? See this section, right above the doc section and the tuto section.
>
>Sure.

Done in https://github.com/artivis/manif/pull/154/commits/c80abaa235e31c18aeeb018c951d4f120190fd03.

examples/se_2_3_localization.cpp
prashanthr05: > We had invented a different Lie group for IMU, see here https://hal.archives-ouvertes.fr/hal-02183498/document, where the time interval is part of the group too.

Really interesting work.

prashanthr05: @artivis @joansola would you have any feedback on this updated example?

joansola: Let me have a look --

prashanthr05:
@joansola I hope I have done a decent improvement for the $SE_2(3)$ localization example in https://github.com/artivis/manif/pull/154/commits/13d837a0af10810ee92694be781a92fbae51df01 with the documentation in terms of setting up the problem and deriving the equations for the filter (mainly system propagation, because the measurement updates remain the same as the one you have suggested in se3_localization example).

I have also fixed the equations to take into account the effect of left Jacobian of $SO(3)$ as per the proper definition of the exponential mapping of $SE_2(3)$.

Please let me know if I have to make any more corrections. Thanks in advance.

The results for constant acceleration and angular velocity motion over a 10 step iteration seems to be like,
``` sh
X STATE     :    X      Y      Z    TH_x   TH_y   TH_z     V_x   V_y    V_z
---------------------------------------------------------------------------
X initial   : +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000
---------------------------------------------------------------------------
X simulated : +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000
X estimated : +0.004 +0.009 -0.002 +0.005 -0.000 -0.005 -0.033 +0.054 -0.044
X unfilterd : +0.000 +0.000 +0.000 +0.060 +0.082 -0.060 -0.033 +0.054 -0.044
---------------------------------------------------------------------------
X simulated : +0.050 +0.000 +0.000 +0.100 +0.010 +0.000 +1.000 +0.000 +0.005
X estimated : +0.054 +0.008 -0.004 +0.099 +0.009 -0.005 +0.969 +0.116 -0.018
X unfilterd : +0.050 -0.001 -0.002 +0.082 +0.193 -0.114 +0.974 +0.090 -0.059
---------------------------------------------------------------------------
X simulated : +0.100 -0.005 +0.001 +0.200 +0.020 +0.000 +2.010 -0.098 +0.015
X estimated : +0.104 -0.001 -0.001 +0.201 +0.018 -0.004 +2.056 -0.009 -0.105
X unfilterd : +0.100 -0.010 -0.007 +0.100 +0.139 -0.086 +2.040 -0.112 -0.267
---------------------------------------------------------------------------
X simulated : +0.151 -0.015 +0.002 +0.300 +0.030 +0.000 +3.029 -0.294 +0.035
X estimated : +0.153 -0.012 -0.002 +0.299 +0.029 -0.000 +3.104 -0.249 -0.051
X unfilterd : +0.150 -0.025 -0.011 +0.245 +0.097 -0.050 +3.069 -0.450 -0.306
---------------------------------------------------------------------------
X simulated : +0.203 -0.029 +0.003 +0.400 +0.040 -0.000 +4.059 -0.588 +0.070
X estimated : +0.204 -0.026 +0.001 +0.401 +0.040 -0.001 +4.222 -0.453 -0.080
X unfilterd : +0.202 -0.037 -0.009 +0.425 +0.106 -0.114 +4.196 -0.617 -0.341
---------------------------------------------------------------------------
X simulated : +0.255 -0.049 +0.006 +0.500 +0.050 -0.000 +5.098 -0.980 +0.124
X estimated : +0.256 -0.045 +0.005 +0.502 +0.049 -0.001 +5.298 -0.885 +0.010
X unfilterd : +0.251 -0.071 +0.002 +0.587 +0.185 -0.014 +5.235 -1.338 -0.085
---------------------------------------------------------------------------
X simulated : +0.308 -0.074 +0.010 +0.600 +0.060 -0.000 +6.147 -1.470 +0.204
X estimated : +0.310 -0.070 +0.008 +0.598 +0.059 -0.002 +6.373 -1.308 +0.019
X unfilterd : +0.301 -0.108 +0.006 +0.742 +0.198 +0.076 +6.246 -2.031 -0.069
---------------------------------------------------------------------------
X simulated : +0.361 -0.103 +0.016 +0.700 +0.070 -0.000 +7.206 -2.059 +0.315
X estimated : +0.362 -0.101 +0.014 +0.701 +0.070 -0.003 +7.450 -1.935 +0.041
X unfilterd : +0.350 -0.145 -0.015 +0.766 +0.082 +0.121 +7.245 -2.813 -0.585
---------------------------------------------------------------------------
X simulated : +0.414 -0.137 +0.023 +0.800 +0.080 -0.000 +8.275 -2.745 +0.461
X estimated : +0.415 -0.137 +0.020 +0.800 +0.078 -0.002 +8.595 -2.608 +0.289
X unfilterd : +0.400 -0.197 +0.006 +0.915 +0.132 +0.228 +8.322 -3.846 -0.054
---------------------------------------------------------------------------
X simulated : +0.468 -0.177 +0.032 +0.900 +0.090 -0.000 +9.353 -3.530 +0.648
X estimated : +0.469 -0.176 +0.030 +0.900 +0.090 +0.000 +9.672 -3.303 +0.473
X unfilterd : +0.456 -0.234 +0.006 +1.048 +0.092 +0.228 +9.434 -4.506 -0.069
---------------------------------------------------------------------------
```

A significant difference between the `X unfiltered` and `X estimated` can already be shown if the iteration count is increased upto 100, when the `X estimated` follows the `X simulated` closely, while `X unfiltered` has already diverged.

A snapshot of such a simulation at the iteration 99 and 100 is as follows,
``` sh
X STATE     :    X      Y      Z    TH_x   TH_y   TH_z     V_x   V_y    V_z
---------------------------------------------------------------------------
X initial   : +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000 +0.000
---------------------------------------------------------------------------
X simulated :   +4.242   +7.052   -2.723   -2.704   -0.270   +0.000  +83.901 +140.987  -54.743
X estimated :   +4.257   +7.036   -2.719   -2.702   -0.265   -0.001  +84.677 +140.621  -54.793
X unfilterd :   +5.344   +7.916   -2.199   -3.014   +0.270   -0.251 +106.587 +158.506  -44.657
---------------------------------------------------------------------------
X simulated :   +4.311   +6.877   -2.974   -2.604   -0.260   +0.000  +85.249 +137.507  -59.753
X estimated :   +4.307   +6.874   -2.979   -2.604   -0.262   +0.000  +85.701 +137.415  -59.978
X unfilterd :   +5.137   +7.816   -2.800   -2.972   +0.152   -0.163 +102.417 +156.539  -56.657
---------------------------------------------------------------------------
```

joansola:
Ill check it out once I'm back from the mountains in 2 days. Thank you!

prashanthr05:
@joansola Have a great time!

examples/se_2_3_localization.cpp
```diff
@@ -86,15 +136,15 @@
...
+ *    U   : control perturbation covariance
  *    b_k : k-th landmark position, R^3
  *    y   : Cartesian landmark measurement in robot frame, R^3
  *    R   : covariance of the measurement noise
  *
  *  The motion and measurement models are
  *
- *    X_(t+1) = f(X_t, u) = X_t * Exp ( w )     // motion equation
+ *    X_(t+1) = f(X_t, u) = X_t * Exp ( u )     // motion equation
  *    [y_k; 1; 0]     = h(X, b_k) = X^-1 * [b_k; 1; 0]          // measurement equation
```
joansola, line 148: We must define the action of SE_3_2. But if this action is on points of R3, then the measurements are points of R3, not of R5:

 b \in R3

 y = X^-1 * b \in R3

mind that in SE3 the action is also on R3: X*b = R*b+T , that is, it is not a homogeneous product but the equivalent Euclidean operation.

examples/se_2_3_localization.cpp
```diff
@@ -24,16 +27,16 @@
...
  *  We assume in this example that the IMU frame coincides with the robot frame.
  *
  *  The robot extended pose X is in SE_2(3) and the beacon positions b_k in R^3,
  *
- *      X = |  R   t  v|              // position, orientation and linear velocity
- *          |  0   1  0|
- *          |  0   0  1|
+ *      X = |    R   p  v|              // position, orientation and linear velocity
+ *          |0 0 0   1  0|
+ *          |0 0 0   0  1|
```
joansola, line 39: I do not think that these leading zeros are necessary, but ok

examples/se_2_3_localization.cpp
```diff
@@ -46,26 +49,73 @@
...
+ * - a is the frame acceleration in mixed-representation (equivalent to p_doubledot in continuous time).
+ * - omega_b as the angular velocity of the robot expressed in the robot frame
+ *
+ * The kinematic equations (1) can be written as,
+ * p <-- p + v dt + 0.5 a dt^2
+ * R <-- R Exp_SO3(omega_b)
+ * v <-- v + a dt
+ *
+ * We would like to express these equations in the form,
+ * X <-- X* Exp(u)
```
joansola, line 70: Use space left and right of `*`

examples/se_2_3_localization.cpp
```diff
@@ -46,26 +49,73 @@
...
+ * We would like to draw a relationship between the sets of equations (1) and (2).
+ * Additionally, we would also like to make use of the IMU measurements as exogeneous inputs.
+ * Considering R.T as the transpose of R, the IMU measurements are modeled as,
+ *    - linear accelerometer measurements alpha = R.T (a - g) + w_acc
+ *    - gyroscope measurements omega = omega_b + w_omega
+ * Note that the IMU measurements are expressed in the IMU frame (coincides with the robot frame - assumption).
+ * The IMU measurements are corrupted by noise,
+ *    - w_omega is the additive white noise affecting the gyroscope measurements
+ *    - w_acc is the additive white noise affecting the linear accelerometer measurements
+ *
```
joansola, line 105: Add a note stating that IMU biases are not considered

examples/se_2_3_localization.cpp
```diff
@@ -46,26 +49,73 @@
...
+ *   u_w = omega dt
+ *   Jinv = inv(JlSO3(u_w)) is the left Jacobian inverse of SO(3) Lie group
+ *   u_p = Jinv * (R.T v dt + 0.5 dt^2 (alpha + R.T g))
+ *   u_v = Jinv * ((alpha + R.T g) dt)
+ *
+ * To cross verify the equations, substituting (3) in (2) will give us (1).
+ * This choice of input vector allows us to directly use measurements from the IMU
+ * for an unified motion integration involving position, orientation and linear velocity of the robot.
+ *
+ * The system propagation noise covariance amtrix becomes,
```
joansola, line 117: *amtrix --> matrix

examples/se_2_3_localization.cpp
```diff
@@ -86,15 +136,15 @@
  *  All these variables are summarized again as follows
  *
  *    X   : robot's extended pose, SE_2(3)
- *    u   : robot control input, (R_{k}.T v dt + 0.5*dt*dt*acc_{k}, omega_{k} dt, acc_{k} dt)
- *    Q   : control perturbation covariance
+ *    u   : robot control input, (Jinv * (R.T v dt + 0.5 dt^2 (alpha + R.T g)), omega dt, Jinv * ((alpha + R.T g) dt))
```
joansola, line 139: I would not repeat here the expression of `u`. Instead, just add ", se_2(3)", akin to the line above

examples/se_2_3_localization.cpp
```diff
@@ -157,8 +214,9 @@ int main()
 
     // IMU measurements in IMU frame
     Vector3d alpha, omega, alpha_prev, omega_prev;
-    alpha << 10.0, 0.0, 9.80665;
-    omega << 1, 0.1, 0;
+    alpha << 10.0, 0.0, 9.80665; // constant acceleration along x-direction
```
joansola, line 217: We should discuss a way to incorporate `g` in a less hardcoded way. We-ll leave this for later though

examples/se_2_3_localization.cpp
joansola: better, use `u = u(X, y) \in se_2_3` with `X` state and `y=[a,w]` IMU readings, and point to expression (number) above in the doc.

examples/se_2_3_localization.cpp
prashanthr05: Done in https://github.com/artivis/manif/pull/154/commits/ded121b75f0f401c518336cf0d31ad862be50275

examples/se_2_3_localization.cpp
prashanthr05: Fixed in https://github.com/artivis/manif/pull/154/commits/ded121b75f0f401c518336cf0d31ad862be50275

examples/se_2_3_localization.cpp
prashanthr05: Fixed in https://github.com/artivis/manif/pull/154/commits/ded121b75f0f401c518336cf0d31ad862be50275

examples/se_2_3_localization.cpp
prashanthr05: Removed in https://github.com/artivis/manif/pull/154/commits/ded121b75f0f401c518336cf0d31ad862be50275

examples/se_2_3_localization.cpp
prashanthr05: Fixed in https://github.com/artivis/manif/pull/154/commits/ded121b75f0f401c518336cf0d31ad862be50275

examples/se_2_3_localization.cpp
prashanthr05: Fixed in https://github.com/artivis/manif/pull/154/commits/ded121b75f0f401c518336cf0d31ad862be50275

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,405 @@
...
+ *    - w_acc is the additive white noise affecting the linear accelerometer measurements
+ * It must be noted that we do not consider IMU biases in the IMU measurement model in this example.
+ *
+ * Taking into account all of the above considerations, the exogenous input vector u (3) becomes,
+ *   u = (u_p, u_w, u_v) where,
+ *   u_w = omega dt
+ *   Jinv = inv(JlSO3(u_w)) is the left Jacobian inverse of SO(3) Lie group
+ *   u_p = Jinv * (R.T v dt + 0.5 dt^2 (alpha + R.T g))
+ *   u_v = Jinv * ((alpha + R.T g) dt)
+ *
```
joansola, line 113: I think here there is some unnecessary confusion.

Let us consider `u = [u_w, u_p, u_v]` angle, position, velocity, in se_2(3).

If one defines the exponential map,

 `Exp(u) = [R(u_w), Jl(u_w)*u_p, Jl(u_w)*u_v ; 0 1 0 ; 0 0 1]`

then the motion integration is simply

 ` X <-- X * Exp(u)`

Now the question is how to define `u`.

If you use Jl.inv in the definition of `u`, then you cancel the effect that these Jacobians have on propagation. This means that you get to the same place as if we have defined the Exp as

 `Exp(u) = [R(u_w), u_p, u_v ; 0 1 0 ; 0 0 1]`            (4)

and having those `Jl` and then those other `Jl.inv` is just pointless.

The issue is that you want to have `u` defined without the Jl.inv:

```
u_w = omega*dt = th
u_p = R.T * v * dt + .5 * (alpha + R.T * g) * dt^2        (5)
u_v = (alpha * R.T * g) * dt
```
leading to the following evolution (5):

```
 p <-- p + Jl * v * dt + 0.5 * Jl * a * dt^2
 R <-- R Exp_SO3(omega_b)                                (6)
 v <-- v + Jl * a * dt
```

which is not equal to (1).

The thing is that this new expression is better than (1) !!! So forget about (1) and do instead:

* compute u as in (5)
* integrate `X <-- X * Exp(u)` as in (2), with Exp() defined as in (4)

Please read Section III.C of the paper we wrote on IMU Lie group for a comment on the differences between (1) and (6) -- it is slightly different here than in the paper, but the concepts are valid.

As I said earlier, both methods converge when dt --> 0. Since we use very small dt, there is not going to be a noticeable difference. But, the approach proposed here is more elegant since it is more accurate and also simpler.

joansola: Notice that things can be done differently. But the key here is that we want to illustrate Lie theory at its best. And this approach here is the cleanest possible.

The approach in (1) considers R3xSO3xR3 instead of SE23, in the sense explored in Example 7 in the Lie paper. On the contrary, (6) is properly a SE23 approach.

joansola: Maybe I made some mistake, especially regarding the `R.T` details in some formulas. If you see things do not fit, please do not hesitate to come back to me for further analysis.

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,405 @@
...
+#include <iostream>
+#include <iomanip>
+
+using std::cout;
+using std::endl;
+
+using namespace Eigen;
+
+typedef Array<double, 3, 1> Array3d;
+typedef Matrix<double, 5, 1> Vector5d;
```
joansola, line 191: This V5 is probably not needed since measurements and beacons are in R3

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,405 @@
...
+    landmarks.push_back(b1);
+    landmarks.push_back(b2);
+    landmarks.push_back(b3);
+    landmarks.push_back(b4);
+
+    // Define the beacon's measurements
+    Vector3d                y, y_noise;
+    Array3d                 y_sigmas;
+    Matrix3d                R;
+    std::vector<Vector5d>   measurements(landmarks.size());
```
joansola, line 260: Use Vector3d since measurements are in R3

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,405 @@
...
+        for (int i = 0; i < landmarks.size(); i++)
+        {
+            b = landmarks[i];                               // lmk coordinates in world frame
+
+            /// simulate noise
+            y_noise = y_sigmas * Array3d::Random();         // measurement noise
+
+            y = X_simulation.inverse().act(b);              // landmark measurement, before adding noise
+            y = y + y_noise;                                // landmark measurement, noisy
+            measurements[i] << y, 1, 0;                     // store for the estimator just below
```
joansola, line 333: remove trailing `1,0` since measurements are R3

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+
+        /// Then we correct using the measurements of each lmk - - - - - - - - -
+        for (int i = 0; i < NUMBER_OF_LMKS_TO_MEASURE; i++)
+        {
+            // landmark
+            b = landmarks[i];                               // lmk coordinates in world frame
+
+           // measurement
+            y = measurements[i].segment<3>(0);                            // lmk measurement, noisy
```
joansola, line 355: Measurements are R3 so no need to do the `segment<>` thing

examples/se_2_3_localization.cpp
prashanthr05: Sorry I missed this.

examples/se_2_3_localization.cpp
prashanthr05: > Please read Section III.C of the paper we wrote on IMU Lie group for a comment on the differences between (1) and (6) -- it is slightly different here than in the paper, but the concepts are valid.

> Notice that things can be done differently. But the key here is that we want to illustrate Lie theory at its best. And this approach here is the cleanest possible.

> The approach in (1) considers R3xSO3xR3 instead of SE23, in the sense explored in Example 7 in the Lie paper. On the contrary, (6) is properly a SE23 approach.

I see. Thanks for noticing this. 

I will go through the formula again and give this a thought to gain some clarity on this and make the necessary changes. 

I shall go through the paper in detail as well.

prashanthr05:
> Hi Prashanth,
> 
> There is a significantly deep comment in this review. I think you will understand the point. I am sorry for being so picky, but the library and the whole Lie effort that we did comes with a purpose, of clarity and exactness, and I believe this purpose is not attained with the PR as it is, because the algebra used in this example is not properly exploiting the SE23 Lie group, but rather the R3xSO3xR3 variant.
> 
> The comment below addresses this issue. It needs however to be verified experimentally if the (small) modifications have the proper effect on the experiment results (I think they will work fine). So, if you can apply these modifications in the code and run the demo again, we can be certain of the method I suggest to truly use SE2(3) Lie group, with the proper Exp() map and the proper tangent space.

@joansola Thanks for being rigorous with review. It helps solidify my understanding as well. I will try to understand and make the required changes.

prashanthr05:
Hi @joansola ,

I have understood that the inclusion of the proper exponential map using the left jacobian of SO(3) actually leads to motion integration equations defined in (6). The cause here is the proper definition of the exponential map and the effect is a more appropriate system propagation equations. Earlier, I had confused the cause and effect inversely. 

I was using proper equations very early on, however my reasoning was wrong which had lead to the mistakes in the documentation. Now I have made the appropriate changes in the description of the example and the code. 

I hope I have been careful enough to avoid other mistakes. Below are the results of the demo you had asked me to run (for different cases).

> So, if you can apply these modifications in the code and run the demo again, we can be certain of the method I suggest to truly use SE2(3) Lie group, with the proper Exp() map and the proper tangent space.

IMU measurements,
```
alpha << 10.0, 0.0, 9.80665; // constant acceleration along x-direction in IMU frme
omega << 1, 0.1, 0; // constant angular velocity about x- and y-direction in IMU frame
```

SE_2(3), dt = 0.1 Iterations 999 and 1000 ( defining u according to (5) )

```
---------------------------------------------------------------------------
X simulated :  +53.530  +13.449  -48.452   -0.232   -0.023   +0.000 +971.892 +261.076 -969.171
X estimated :  +53.452  +13.401  -48.475   -0.240   -0.024   +0.007 +970.611 +260.576 -969.686
X unfilterd :  +44.288  -30.206   +3.034   +1.623   +0.233   +1.121 +807.512 -613.171  +23.582
---------------------------------------------------------------------------
X simulated :  +53.833  +11.021  -48.730   -0.132   -0.013   +0.000 +977.751 +212.487 -974.631
X estimated :  +53.750  +10.964  -48.758   -0.152   -0.016   +0.018 +976.403 +212.365 -975.200
X unfilterd :  +44.810  -28.948   +4.339   +1.622   +0.295   +1.088 +819.312 -590.440  +47.851
---------------------------------------------------------------------------
```



R3 x SO3 x R3, dt = 0.1 Iterations 999 and 1000 (canceling the effects of  Jl using Jlinv in input vector u)

```
---------------------------------------------------------------------------
X simulated :  +53.773  +11.021  -48.811   -0.232   -0.023   +0.000 +976.752 +212.481 -976.549
X estimated :  +53.691  +10.970  -48.837   -0.241   -0.024   +0.008 +975.388 +211.953 -977.132
X unfilterd :  +43.469  -30.797   +4.224   +1.623   +0.233   +1.121 +791.231 -624.930  +47.060
---------------------------------------------------------------------------
X simulated :  +54.077   +8.580  -48.966   -0.132   -0.013   +0.000 +982.634 +163.660 -979.547
X estimated :  +53.991   +8.521  -48.994   -0.153   -0.015   +0.019 +981.224 +163.504 -980.182
X unfilterd :  +43.992  -29.549   +5.518   +1.622   +0.295   +1.088 +803.038 -602.389  +71.126
---------------------------------------------------------------------------
```





SE_2(3) ,dt = 0.001 Iterations 999 and 1000 ( defining u according to (5) )

```
---------------------------------------------------------------------------
X simulated :  +0.006  -0.003  +0.000  +0.998  +0.100  -0.000 +10.468  -4.879  +0.832
X estimated :  +0.007  -0.002  +0.001  +0.997  +0.101  +0.002 +10.950  -6.539  +0.483
X unfilterd : +0.003 +0.003 +0.003 -2.651 -0.013 -1.611 +9.437 +5.491 +7.488
---------------------------------------------------------------------------
X simulated :  +0.006  -0.003  +0.000  +0.999  +0.100  -0.000 +10.479  -4.889  +0.835
X estimated :  +0.007  -0.002  +0.001  +1.003  +0.100  +0.001 +11.032  -6.523  +0.471
X unfilterd :  +0.004  -0.003  +0.001  +2.638  +0.024  +1.604 +11.189  -5.149  +4.842
---------------------------------------------------------------------------
```



R3 x SO3 x R3, dt = 0.001 Iterations 999 and 1000 (canceling the effects of  Jl using Jlinv in input vector u)

```
---------------------------------------------------------------------------
X simulated :  +0.006  -0.003  +0.000  +0.998  +0.100  -0.000 +10.468  -4.879  +0.835
X estimated :  +0.007  -0.002  +0.001  +0.997  +0.101  +0.002 +10.950  -6.539  +0.486
X unfilterd : +0.003 +0.003 +0.003 -2.651 -0.013 -1.611 +9.438 +5.490 +7.486
---------------------------------------------------------------------------
X simulated :  +0.006  -0.003  +0.000  +0.999  +0.100  -0.000 +10.479  -4.889  +0.838
X estimated :  +0.007  -0.002  +0.001  +1.003  +0.100  +0.001 +11.032  -6.522  +0.473
X unfilterd :  +0.004  -0.003  +0.001  +2.638  +0.024  +1.604 +11.187  -5.148  +4.843
---------------------------------------------------------------------------
```

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+
+
+
+    // START TEMPORAL LOOP
+    //
+    //
+
+    // Make 10 steps. Measure up to three landmarks each time.
+    for (int t = 0; t < 10; t++)
```
joansola, line 307: I think `t` and `dt` should be consistent. Here, `t` is now a counter, thus without units, and `dt` is in seconds. I suggest replacing this for() by
```
for (double t = 0; t < 10; t += dt)
```

joansola: the `10` can be changed by whatever suits you best

prashanthr05:
> We're almost there. Thanks for the effort and patience.

I must thank you for your patience and time.

> Check out the comment below on the for() loop and see if fixing all simulations with the same final time (e.g. t=10) yields a better set of experiments to evaluate.

All of the following experiments were done for `t (begin) = 0` to `t (final) = 10`, for different sampling periods with the counter,
```
for (double t = 0; t < 10; t+=dt)
```
In the end, there is a table enlisting the norm of the errors between the `SE_2(3)` and `R3 x SO(3) x R3` state coefficients for different sampling periods (separated in log10 scale 0.1 to 0.0001).


SE_2(3), dt = 0.1,  t (final) = 10 ( defining u according to (5) )

```
---------------------------------------------------------------------------
X simulated :   +4.297   +7.014   -2.647   -2.604   -0.260   +0.000  +84.972 +140.278  -53.208
X estimated :   +4.292   +7.012   -2.654   -2.604   -0.263   +0.000  +85.394 +140.196  -53.461
X unfilterd :   +5.136   +7.957   -2.437   -2.972   +0.152   -0.163 +102.422 +159.383  -49.388
---------------------------------------------------------------------------
X simulated :   +4.365   +6.847   -2.896   -2.504   -0.250   +0.000  +86.305 +136.945  -58.167
X estimated :   +4.368   +6.845   -2.891   -2.504   -0.249   -0.002  +86.901 +136.908  -58.300
X unfilterd :   +5.450   +7.844   -2.323   -2.900   +0.252   -0.222 +108.718 +157.157  -47.253
---------------------------------------------------------------------------
```



R3 x SO3 x R3, dt = 0.1, t (final) = 10 (canceling the effects of Jl using Jlinv in input vector u)

```
---------------------------------------------------------------------------
X simulated :   +4.311   +6.877   -2.974   -2.604   -0.260   +0.000  +85.249 +137.507  -59.753
X estimated :   +4.307   +6.874   -2.979   -2.604   -0.262   +0.000  +85.701 +137.415  -59.978
X unfilterd :   +5.137   +7.816   -2.800   -2.972   +0.152   -0.163 +102.417 +156.539  -56.657
---------------------------------------------------------------------------
X simulated :   +4.380   +6.698   -3.213   -2.504   -0.250   +0.000  +86.607 +133.930  -64.533
X estimated :   +4.384   +6.696   -3.208   -2.504   -0.248   -0.002  +87.215 +133.897  -64.656
X unfilterd :   +5.462   +7.693   -2.681   -2.900   +0.252   -0.222 +108.940 +154.112  -54.436
---------------------------------------------------------------------------
```

SE_2(3), dt = 0.01,  t (final) = 10 ( defining u according to (5) )

```
---------------------------------------------------------------------------
X simulated :   +0.481   +0.679   -0.303   -2.514   -0.251   +0.000  +86.328 +135.721  -63.338
X estimated :   +0.483   +0.679   -0.303   -2.517   -0.251   +0.000  +86.231 +140.217  -62.516
X unfilterd :   +0.195   +0.584   -0.274   -0.628   -1.205   -1.319  +38.800 +116.734  -58.988
---------------------------------------------------------------------------
X simulated :   +0.482   +0.677   -0.306   -2.504   -0.250   +0.000  +86.464 +135.357  -63.808
X estimated :   +0.484   +0.676   -0.306   -2.505   -0.251   +0.002  +86.385 +139.761  -63.169
X unfilterd :   +0.221   +0.569   -0.265   -0.643   -1.159   -1.244  +43.992 +113.732  -56.983
---------------------------------------------------------------------------
```

R3 x SO3 x R3, dt = 0.01, t (final) = 10 (canceling the effects of Jl using Jlinv in input vector u)

```
---------------------------------------------------------------------------
X simulated :   +0.482   +0.678   -0.307   -2.514   -0.251   +0.000  +86.360 +135.404  -63.973
X estimated :   +0.483   +0.678   -0.306   -2.517   -0.251   +0.000  +86.262 +139.900  -63.151
X unfilterd :   +0.198   +0.584   -0.275   -0.628   -1.205   -1.319  +39.290 +116.672  -59.061
---------------------------------------------------------------------------
X simulated :   +0.482   +0.676   -0.309   -2.504   -0.250   +0.000  +86.496 +135.037  -64.441
X estimated :   +0.484   +0.675   -0.309   -2.505   -0.251   +0.002  +86.416 +139.441  -63.803
X unfilterd :   +0.223   +0.568   -0.265   -0.643   -1.159   -1.244  +44.474 +113.651  -57.047
---------------------------------------------------------------------------
```



SE_2(3), dt = 0.001,  t (final) = 10 ( defining u according to (5) )

```
---------------------------------------------------------------------------
X simulated :   +0.093   +0.068   -0.019   -2.505   -0.251   -0.000  +86.468 +135.216  -64.320
X estimated :   +0.094   +0.068   -0.019   -2.505   -0.250   -0.002  +83.823 +137.383  -53.766
X unfilterd :  -0.011  -0.031  -0.013  -2.764  -0.477  -0.126 +15.718 -21.753 +21.059
---------------------------------------------------------------------------
X simulated :   +0.093   +0.068   -0.019   -2.504   -0.250   -0.000  +86.482 +135.180  -64.367
X estimated :   +0.095   +0.068   -0.019   -2.503   -0.248   -0.001  +83.992 +137.164  -53.866
X unfilterd :  -0.012  -0.031  -0.012  -2.760  -0.507  -0.031 +16.272 -22.169 +20.517
---------------------------------------------------------------------------
```



R3 x SO3 x R3, dt = 0.001, t (final) = 10 (canceling the effects of Jl using Jlinv in input vector u)

```
---------------------------------------------------------------------------
X simulated :   +0.093   +0.068   -0.019   -2.505   -0.251   -0.000  +86.472 +135.184  -64.383
X estimated :   +0.094   +0.068   -0.019   -2.505   -0.250   -0.002  +83.826 +137.350  -53.829
X unfilterd :  -0.011  -0.031  -0.013  -2.764  -0.477  -0.126 +15.721 -21.773 +21.069
---------------------------------------------------------------------------
X simulated :   +0.093   +0.068   -0.019   -2.504   -0.250   -0.000  +86.485 +135.147  -64.430
X estimated :   +0.095   +0.068   -0.019   -2.503   -0.248   -0.001  +83.995 +137.131  -53.929
X unfilterd :  -0.012  -0.031  -0.012  -2.760  -0.507  -0.031 +16.276 -22.190 +20.527
---------------------------------------------------------------------------
```



SE_2(3), dt = 0.0001,  t (final) = 10 ( defining u according to (5) )

```
---------------------------------------------------------------------------
X simulated :   +0.054   +0.007   +0.010   -2.504   -0.250   -0.000  +86.482 +135.165  -64.418
X estimated :   +0.056   +0.007   +0.010   -2.501   -0.249   -0.002  +78.470 +136.807  -78.142
X unfilterd :  -0.000  -0.005  -0.003  -1.499  -0.215  -0.212  +9.712 +14.881 +13.247
---------------------------------------------------------------------------
X simulated :   +0.054   +0.007   +0.010   -2.504   -0.250   -0.000  +86.484 +135.162  -64.423
X estimated :   +0.056   +0.007   +0.010   -2.501   -0.247   +0.001  +78.681 +136.690  -78.198
X unfilterd :  -0.000  -0.005  -0.002  -1.432  -0.274  -0.167 +10.264 +15.146 +12.239
---------------------------------------------------------------------------
```

R3 x SO3 x R3, dt = 0.0001, t (final) = 10 (canceling the effects of Jl using Jlinv in input vector u)

```
---------------------------------------------------------------------------
X simulated :   +0.054   +0.007   +0.010   -2.504   -0.250   -0.000  +86.483 +135.162  -64.424
X estimated :   +0.056   +0.007   +0.010   -2.501   -0.249   -0.002  +78.471 +136.804  -78.148
X unfilterd :  -0.000  -0.005  -0.003  -1.499  -0.215  -0.212  +9.712 +14.881 +13.247
---------------------------------------------------------------------------
X simulated :   +0.054   +0.007   +0.010   -2.504   -0.250   -0.000  +86.484 +135.159  -64.429
X estimated :   +0.056   +0.007   +0.010   -2.501   -0.247   +0.001  +78.681 +136.687  -78.205
X unfilterd :  -0.000  -0.005  -0.002  -1.432  -0.274  -0.167 +10.264 +15.146 +12.239
---------------------------------------------------------------------------
```



The norm of difference in the simulated state coefficients for different sampling periods `dt` are,

| dt   (s)  | norm(x__r3so3r3_sim - x_sim_se23) |
| ------ | --------------------------------- |
| 0.0001 | 0.0067                            |
| 0.001  | 0.0712                            |
| 0.01   | 0.7100                            |
| 0.1    | 7.0591                            |

artivis:
@prashanthr05 Out of curiosity, how do you compute the `norm` in your above results (`norm(x__r3so3r3_sim - x_sim_se23)`)?? You may want to have a look at the `weightedNorm` provided by the Tangent class (`(x__r3so3r3_sim - x_sim_se23).weightedNorm()`). That should **not** change the trend highlighted by your table but weight e.g. the rotation contribution vs translation. So, no need to re-run everything ;)

prashanthr05:
> @prashanthr05 Out of curiosity, how do you compute the `norm` in your above results (`norm(x__r3so3r3_sim - x_sim_se23)`)?? You may want to have a look at the `weightedNorm` provided by the Tangent class (`(x__r3so3r3_sim - x_sim_se23).weightedNorm()`). That should **not** change the trend highlighted by your table but weight e.g. the rotation contribution vs translation. So, no need to re-run everything ;)

Ciao @artivis ! I just used the regular vector norm from Matlab for the log() outputs. But, I see that the weighted norm uses a `W()` method which in turn uses a `run()` method. I thought `run()` was used to get the i-th generator for the Lie algebra. It's not clear to me how this is being used in the weighted norm. Could you explain its working?

artivis:
>  I see that the weighted norm uses a W() method which in turn uses a run() method. I thought run() was used to get the i-th generator for the Lie algebra. It's not clear to me how this is being used in the weighted norm. Could you explain its working?

Humm that's really implementation details but it is a design that I'm planning to generalize in `manif` in the future. As you pointed, `W()` [calls the `run` function](https://github.com/artivis/manif/blob/b1c82befd066fb4e954742bab56c7795168bdaab/include/manif/impl/tangent_base.h#L671) from [the class (or traits) `WEvaluator`](https://github.com/artivis/manif/blob/b1c82befd066fb4e954742bab56c7795168bdaab/include/manif/impl/generator.h#L20). Mind that it is **not** [the class `GeneratorEvaluator`](https://github.com/artivis/manif/blob/b1c82befd066fb4e954742bab56c7795168bdaab/include/manif/impl/generator.h#L8), I believe your confusion comes from there (`WEvaluator::run` != `GeneratorEvaluator::run`).
This class (`WEvaluator`) automatically [computes a weight matrix `W` from the group basis](https://github.com/artivis/manif/blob/b1c82befd066fb4e954742bab56c7795168bdaab/include/manif/impl/generator.h#L27) which in turn is used by the function `weightedNorm` to compute the Euclidean weighted norm.

prashanthr05:
>  Mind that it is not the class GeneratorEvaluator, I believe your confusion comes from there (WEvaluator::run != GeneratorEvaluator::run).

Indeed. Now, I understood the flow. Thanks for the explanation!

prashanthr05:
Ciao @joansola ! Would you have any feedback on the latest tests that I had run after your advice? Please see https://github.com/artivis/manif/pull/154#issuecomment-675541556

joansola:
Sorry @prashanthr05 I have been out this last week! Let me have a look.

joansola:
OK @prashanthr05 I have a last comment regarding this demo. It relates to the way measurements are created.

You have created a constant acc measurement of `alpha << 10.0, 0.0, 9.80665;` I have a couple of comments that might help make this demo a little more realistic.

1. The first and most important relates to the direction of gravity. Since gravity is in the global frame and alpha is local, this `9.80655` will not compensate except for the case where the IMU is absolutely flat, which only happens at t=0.

    Basically, you have created `alpha = (10,0,0) - g`

    I would do things different:

    - Set an acceleration alpha in local frame, without gravity compensation. This can be constant, e.g. `alpha = (10, 0, 0)`
    - Add gravity from the global frame, with `alpha -= R * g`. Here importantly `R` is the rotation matrix OF THE SIMULATOR, that is, the true IMU orientation.

    The result is `alpha = (10,0,0) - R*g`

    Notice that this alpha needs to be updated at each step of the `for()` loop. This is of course no problem.


2. Make the real acceleration smaller so that in 10 seconds the IMU does not fly away that far. This means that the landmarks, which are nearby, will still have a noticeable impact on observability. Otherwise, if you go too far from the landmarks, their observation is not going to provide much info the the robot localization.

    I would use values in the range (0.5 -- 2) m/s2.

    It'd be interesting to check the final position of the IMU at t=10, and see where it is with respect to the landmarks. If it is too far away, try to play with the control values alpha and omega to reach a nice final position. You can also, of course, change the landmark positions. But I would not do it: I would keep the same positions as in the SE3 example, so that comparisons can be made if desired.

With both 1) and 2) together, you'd get something like `alpha = (2,0,0) - R*g`.

Both modifications will make this demo more realistic and therefore more useful.

joansola:
It seems to me these would be the last modifications requested. Let us see what comes out of it. I guess the PR could be merged soon after. @artivis what do you think?

joansola:
I think plots of the trajectories (only XYZ positions) would help evaluate this demo better. I am surprised that after 10 seconds in your last demos, the values of position are so small!! `X simulated :   +0.054   +0.007   +0.010 ...`

This small final displacement, with acc_x = 10.0 and without a proper gravity compensation, during 10 seconds, is very unlikely in my opinion.

prashanthr05:
First of all, thank you for your feedback. And I am sorry, I haven't been really proactive with the demo code. I appreciate your suggestions and I will do the necessary changes and try to add plots for the same in the coming days.

> Both modifications will make this demo more realistic and therefore more useful.

I agree with this. Also maybe we can test for another scenario were also the angular velocities are non-zero. 

> I think plots of the trajectories (only XYZ positions) would help evaluate this demo better. I am surprised that after 10 seconds in your last demos, the values of position are so small!! `X simulated : +0.054 +0.007 +0.010 ...`
> 
> This small final displacement, with acc_x = 10.0 and without a proper gravity compensation, during 10 seconds, is very unlikely in my opinion. Even more if the final velocities are so high! `X simulated : ... +86.483 +135.162 -64.424`
> 
> Am I interpreting the results wrongly? I think we have `X,Y,Z-roll,pitch,yaw-Vx,Vy,Vz` no? Maybe printing the `X.log()` values is not the way to go? You can try printing `X.translation() ; X.rotation().log() ; X.velocity()` instead, I think it makes more sense in order to interpret the results.

Oops! I totally overlooked the fact about the positions being small, while the velocities are so high. I will try to repeat the experiments and try to plot the quantities to have a better picture. And maybe also for the orientation, plotting the Euler angles would be more intuitive than looking at the log(), I suppose? 

I apologize again for overlooking/not being critical about the outputs.

joansola:
> I agree with this. Also maybe we can test for another scenario were also the angular velocities are non-zero.

Angular velocities are already non-zero, no need to change this. This is why the IMU rotates and this is why gravity is not compensated away from t=0.
> 

> Oops! I totally overlooked the fact about the positions being small, while the velocities are so high. I will try to repeat the experiments and try to plot the quantities to have a better picture. And maybe also for the orientation, plotting the Euler angles would be more intuitive than looking at the log(), I suppose?
> 
> I apologize again for overlooking/not being critical about the outputs.

Best is to write a `log.txt` of the full trajectory of the full state, then in Matlab or elsewhere plot the trajectories. I also used to have a 3D visualizer where I could animate the real motion of the body. With such tools you can verify if results fit with the intended trajectories you are producing with the IMU measurements.

It seems the estimator works since X_simu ~ X_est. However, it could be that some motion formula is wrong. Since you use the same formulae for simu and for estimation, they match, but they do not seem to fit well with the expected motion.

Another thing to do is also go back to your original formulation (before all mathematical changes that I have been suggesting) see if it worked before, and therefore see if we have broken anything on the way.

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,412 @@
...
+    landmarks.push_back(b4);
+
+    // Define the beacon's measurements
+    Vector3d                y, y_noise;
+    Array3d                 y_sigmas;
+    Matrix3d                R;
+    std::vector<Vector3d>   measurements(landmarks.size());
+
+    y_sigmas << 0.01, 0.01, 0.01;
+    R        = (y_sigmas * y_sigmas).matrix().asDiagonal();
```
joansola, line 272: OK so R is the covariances matrix of the IMU measurements.

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,412 @@
...
+    {
+        //// I. Simulation ###############################################################################
+
+        /// get current state and measurements from previous step
+        auto R_k = X_simulation.rotation();
+        auto v_k = X_simulation.linearVelocity();
+        auto acc_k = alpha_prev + R_k.transpose()*g;
+
+        /// input vector
+        u_nom << (dt*(R.transpose())*v_k + 0.5*dt*dt*acc_k),  dt*omega_prev, dt*acc_k;
```
joansola, line 317: Here you use `R` as the rotation matrix! But `R` is the covariances matrix! Since you are in the simulator part, you must use `R_k` here and not `R`.

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+        /// simulate noise
+        u_noise = u_sigmas * Array9d::Random();             // control noise
+        u_noisy = u_nom + u_noise;                          // noisy control
```
joansola, line 321: Since you used the simulated `R_k` to compute `u_nom`, then `u_nom` is only valid for the simulator.

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+        u_simu   = u_nom;
+        u_est    = u_noisy;
```
joansola, line 324: This means that `u_est` cannot be computed from `u_nom`. You need to re-compute a new `u` using the estimated rotation matrix. Than add noise to create `u_est`.

joansola (commented):
I detected a number of bugs related to the way you compute the control vector `u`.

- the first and most important is that you used a wrong matrix `R` which is not a rotation matrix but a covariance matrix!!
- the second bug is that you use the same rotation in the simulated and estimated `u`. You need to do a completely independent computation of `u` for the simulator and the estimator, using `R_simu` for one, and `R_est` for the other.

Look at the comments in the file for details.

examples/se_2_3_localization.cpp
prashanthr05: Thank you for identifying this! That was a silly mistake from my part.

prashanthr05:
Ciao @joansola @artivis !! 

I had made the suggested changes to the localization demo. Mainly,
- Fix the bug in the mismatch between measurement covariance matrix and rotation `R`
- Fix the usage of proper control input for the estimation which is a function of current state estimate and not the simulated state
- Use a constant free body acceleration in the IMU frame which is compensated by gravity at every time step by rotating it into the body frame using the simulated rotation.
- Check the estimator outputs for the position, roll-pitch-yaw, and velocity instead of checking for `X.log()`.

However, looks like the estimator is failing in general. The divergence between the estimates and the simulated outputs seem to be really high. 

For faster debugging, I replicated the example in Matlab [here](https://github.com/prashanthr05/coding_playground/blob/master/manif-examples-test-matlab/se2_3localization.m) and compare it with the C++ results copied into a file.

|3D position | Estimates |
|:---:|:---:|
|se23pos|se23estimates|



As a sanity check, I also replicated the SE(3) localization example (checking for both Matlab and C++ implementation). This seems to be working fine.
|3D position | Estimates |
|:---:|:---:|
|se3pos|se3estimates|


I would need more time to debug the SE2(3) localization demo. However, I am overloaded with tasks at the moment and I am not sure how sooner I can do the debugging.

In case, if this PR is blocking #147 I would suggest proceeding with that PR, and I will make the relevant changes within this PR to be compatible with the newer changes introduced in #147 .

Kindly let me know what you think.

joansola:
thanks @prashanthr05 . I also agree on merging if blocking. Maybe add a WARNING at the top of the demo file, or better still exclude it from the merge while it is buggy. We-ll fix it soon.

joansola:
You can immediately open a new PR with the demo file and we'll keep working on that.

Waiting for @artivis to say his word.

artivis:
Hi,
so I'd propose to merge #147 and keep this PR opened until it is fixed. Especially since we're not sure where does the issue come from. Updating this PR to post #147 merge won't be difficult anyway.

prashanthr05:
> Hi,
> so I'd propose to merge #147 and keep this PR opened until it is fixed. Especially since we're not sure where does the issue come from. Updating this PR to post #147 merge won't be difficult anyway.

Ok for me! I will update this PR to post #147 after it is merged.

joansola:
What's the status of this?

artivis:
> What's the status of this?

Looking at the backlog it seems that it requiers some debugging

joansola:
> > What's the status of this?
> 
> Looking at the backlog it seems that it requiers some debugging

Indeed. My message was just a ping to wake up the subject.

prashanthr05:
> > > What's the status of this?
> > 
> > 
> > Looking at the backlog it seems that it requiers some debugging
> 
> Indeed. My message was just a ping to wake up the subject.

Ciao @joansola, @artivis !!
Yes, Thank you for the wake-up call :D !!.
I have been busy with some deadlines this month. 
I shall work on the demo this weekend and try to debug the issue of divergence in the se_2_3_localization example.

prashanthr05:
Apparently, there was a bug in the state covariance propagation,

where I was using 

$$ P = J_x P  J_x^T + J_u U J_u^T $$
where, $ J_x =  Adj (Exp(-u)), J_u = Jr(-u)$, and $Jr$ is the right Jacobian of the matrix Lie group. 

I changed this to,
$$ P = F P F^T + J_u U J_u^T $$
where  $F = J_x + J_u C $ and C is the Jacobian of $u(X Exp(\epsilon))$ with respect to $\epsilon$. $\epsilon$ is a small perturbation in the state vector. Reference paper: [Discrete Extended Kalman Filter on Lie Groups](https://hal.archives-ouvertes.fr/hal-00903252/document).

This change is done in https://github.com/artivis/manif/pull/154/commits/b44d6355ba53216ece55a3345f8eaa76e3c15490.



Terminal state for an experiment of duration 10 seconds with sample period 0.01 seconds.
```
---------------------------------------------------------------------------
---------------------------------------------------------------------------
X simulated : Pos: +6.184 +0.382 +3.028 RPY: +0.155 +0.996 +0.084 Vel: +1.302 +0.070 +0.387
X estimated : Pos: +6.182 +0.383 +3.028 RPY: +0.160 +0.996 +0.090 Vel: +1.302 +0.127 +0.363
X unfilterd : Pos: +5.840 +0.094 +3.560 RPY: +0.444 +0.967 +0.185 Vel: +1.335 -0.049 +0.443
---------------------------------------------------------------------------
X simulated : log: +4.152 +0.585 +5.839 +0.100 +0.999 +0.000 +0.999 +0.100 +1.000
X estimated : log: +4.150 +0.580 +5.839 +0.101 +0.999 +0.003 +1.011 +0.154 +0.976
X unfilterd : log: +3.609 +0.939 +6.082 +0.319 +0.982 -0.047 +1.007 +0.087 +1.064
---------------------------------------------------------------------------
---------------------------------------------------------------------------
X simulated : Pos: +6.197 +0.382 +3.031 RPY: +0.155 +0.997 +0.085 Vel: +1.303 +0.070 +0.386
X estimated : Pos: +6.197 +0.383 +3.031 RPY: +0.156 +0.997 +0.086 Vel: +1.316 +0.121 +0.362
X unfilterd : Pos: +5.853 +0.092 +3.564 RPY: +0.458 +0.966 +0.200 Vel: +1.341 -0.048 +0.446
---------------------------------------------------------------------------
X simulated : log: +4.159 +0.586 +5.851 +0.100 +1.000 +0.000 +1.000 +0.100 +1.001
X estimated : log: +4.159 +0.585 +5.851 +0.100 +1.000 +0.000 +1.024 +0.150 +0.983
X unfilterd : log: +3.616 +0.929 +6.096 +0.324 +0.984 -0.039 +1.011 +0.086 +1.071
---------------------------------------------------------------------------
---------------------------------------------------------------------------
```

Plotting matlab vs cpp implementation
|xyz plot| state time - evolution |
|:---:|:---:|
|se23dlgekflocalization|se23estimates|

The Matlab implementation can be found in https://github.com/prashanthr05/coding_playground/blob/master/manif-examples-test-matlab/se2_3imuDLGEKF.m

joansola:
Hi, this looks good! Thanks.

I-ll have a look at the final code tomorrow.

artivis:
@joansola Any update? Are we good to merge?

joansola:
Hi. I haven't had time to really look into it. There are two things to check a-priori (I mean, before I even look at the code):

- The necessity of applying the skew operator makes me think whether the Jacobians are applied properly or not (I repeat I have not looked at the code, it's just a warning for me)
- I would like the Jacobians to have consistent names following some kind of rule. I like `J_y_x := dy/dx` so that you can build complex Jacobians with the chain rule, `J_z_x = J_z_y*J_y_x`

Both things are waiting until next week when I'll be free from my exams at CSIC. In the meantime, and if this MR is blocking, maybe @prashanthr05  you can comment on the two topics above.

artivis:
> Both things are waiting until next week when I'll be free from my exams at CSIC. In the meantime, and if this MR is blocking, maybe @prashanthr05 you can comment on the two topics above.

Right, don't worry about this it can wait for you to be done with the exams :+1:

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,452 @@
...
+    manif::SE_2_3Tangentd  u_simu, u_est, u_unfilt;
+    Vector9d            u_nom, u_noisy, u_noise;
+    Array9d             u_sigmas;
+    Matrix9d            U;
+
+    u_sigmas << 0.0, 0.0, 0.0, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01;
+    U        = (u_sigmas * u_sigmas).matrix().asDiagonal();
+
+    // Declare the Jacobians of the motion wrt robot and control
+    manif::SE_2_3d::Jacobian slantFk, Fk, J_x, J_u;
```
joansola, line 250: Use `J_y_x = dy/dx` consistently for each one of the functions. then use the chain rule as in e.g. `J_z_x = J_z_y*J_y_x` or even `J_z_x = J_z_y*J_y_x + J_z_u*J_u_x`.

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,452 @@
...
+    // Declare the Jacobian of the measurements wrt the robot pose
+    Matrix<double, 3, 9>    H;      // H = J_e_x
+
+    // Declare some temporaries
+    Vector3d                e, z;   // expectation, innovation
+    Matrix3d                E, Z;   // covariances of the above
+    Matrix<double, 9, 3>    K;      // Kalman gain
+    manif::SE_2_3Tangentd      dx;     // optimal update step, or error-state
+    manif::SE_2_3d::Jacobian   J_xi_x; // Jacobian is typedef Matrix
+    Matrix<double, 3, 9>       J_e_xi; // Jacobian
```
joansola, line 284: See that these Jacobians already have proper names J_y_x = dy/dx

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+
+
+
+        //// II. Estimation ###############################################################################
+
+        /// get current state estimate to build the state-dependent control vector
+        auto R_k_est = X.rotation();
+        auto v_k_est = X.linearVelocity();
+        auto acc_k_est = alpha_prev + R_k_est.transpose()*g;
```
joansola, line 359: Can this be done with X.act() ? Or maybe X.inv().act()?

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+        Eigen::Vector3d accLin = dt*(R_k_est.transpose())*v_k_est + 0.5*dt*dt*acc_k_est;
+        Eigen::Vector3d gLin = R_k_est.transpose()*g*dt;
+        Eigen::Matrix3d accLinCross = manif::skew(accLin);
+        Eigen::Matrix3d gCross = manif::skew(gLin);
```
joansola, line 364: Same for these lines. Can we use the Lie group operators directly on the object X and not breaking it into pieces?

joansola (commented):
I have raised one small and one possibly big issue:

- small: Names of Jacobians. Since we have a few intricate operations, I suggest chosing better names for Jacobians 
- big: use of the manif tools for the SE2(3) group instead of breaking X\in SE2(3) into its constituents SO3, R3 and R3. This second one I am not sure but it is worth the effort, so if doable, this file should be doing it to serve as a proper tutorial.

examples/se_2_3_localization.cpp
prashanthr05: The X.act() will cause the position to be added to this equation.

examples/se_2_3_localization.cpp
prashanthr05: Since this is the Jacobian of a state-dependent input motion vector with respect to a state perturbation vector, we do not already have some methods that would give us this Jacobian. 

Which is why I had split it into these equations.
It also depends on the input vector we might define. (The above is specific for a strap down IMU based motion model).

examples/se_2_3_localization.cpp
joansola: OK

examples/se_2_3_localization.cpp
joansola: I think doing these easy changes will do. I am checking below for the Jacobian of the input vector to see if we can also improve the notation.

examples/se_2_3_localization.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+        X = X.plus(u_est, J_x, J_u);                        // X * exp(u), with Jacobians
+
+        // Prepare Jacobian
+        slantFk.setZero();
+        slantFk.block<3, 3>(0, 3) = accLinCross;
+        slantFk.block<3, 3>(0, 6) = Eigen::Matrix3d::Identity()*dt;
+        slantFk.block<3, 3>(6, 3) = gCross;
+        Fk = J_x + J_u*slantFk;
+
```
joansola, line 379: Rename :
- `Fk`: leave as is if you want
- `J_x` --> `J_x_x` which is df/dx
- `J_u` --> `J_x_u`
- `slantFk` --> `J_u_x`
This results in L378 being:
- `Fk = J_x_x + J_x_u*J_u_x` which used the chain rule

examples/se_2_3_localization.cpp
joansola: OK. Check my comment below.

joansola:
Hi @prashanthr05  with the minor changes suggested above (Jacobian renaming) this PR will be ready!!!

examples/se_2_3_localization.cpp
prashanthr05: Thank you, @joansola,  for the suggestion! These names are rather quite informative than the former ones.
I have fixed this in https://github.com/artivis/manif/pull/154/commits/b00e27afe04fdf84e7fb052a6d796614d8da83b4.

joansola: Yea that's the point! More information, easy to understand --> less errors!

joansola (approved):

joansola:
Thanks @prashanthr05 you did a fine job!

Let us wait for CI and let us have @artivis do the merge when he considers.

prashanthr05:
> Thanks @prashanthr05 you did a fine job!

Thanks to you, @joansola !! I appreciate your thorough review! 

> Let us wait for CI and let us have @artivis do the merge when he considers.

Sure.

artivis:
Thanks to both of you @prashanthr05, @joansola.
This is a great addition and a super clean PR. I'll re-run the failing test (seems unrelated), once all green, I'll merge!

artivis (approved):

prashanthr05:
> Thanks to both of you @prashanthr05, @joansola.
> This is a great addition and a super clean PR. I'll re-run the failing test (seems unrelated), once all green, I'll merge!

Thank you @artivis !!

PR #155, opened 2020-08-04
Add option to limit tests

artivis:
Compiling the `Rn` test suite often hangs in CI on arm64 - it is a rather heavy suite... This PR introduces an option to limit the `Rn` tests to a bare minimum and enables it in the arm64 job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -289,7 +289,7 @@ matrix:
     arch: arm64
     compiler: gcc
     script:
-      - cmake -DBUILD_EXAMPLES=ON -DBUILD_TESTING=ON ..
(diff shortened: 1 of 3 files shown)


artivis:
> I guess this is a workaround... OK

It is, but it is only enabled for that one job (arm64). Other jobs are not affected by this.

artivis:
There seems to be way to [increase Travis timeouts](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) including the one we're concerned with here (no logs for more than 10mins).
I'll give it a shot in another PR (replacing this one).

artivis:
@joansola I'm closing this PR in favor of #158.

PR #165, opened 2020-08-18
Add unit dual quaternion

### LINKED ISSUE #164: Interest in Unit Dual Quaternions?
After seeing [this post on ROS Dicourse](https://discourse.ros.org/t/release-of-dual-quaternions/15911) and the comments asking for a c++ implementation, I wondered if 'Unit Dual Quaternions' (`DH_u` I believe) would be an interesting/useful addition to `manif` given that it is somewhat equivalent to `SE3`.

artivis:
This PR adds the unit dual quaternion group,  currently called `DHu`.

Draft implementation:

+ [x] `DHu::Random`
+ [x] `DHu::Inverse`
+ [x] `DHu::Compose`
+ [x] `DHu::normalize`
+ [x] `DHu::normalize`
+ [x] `DHu::normalize`

Comes on top of #150.

Closes #164.

### DIFF
--- a/include/manif/impl/dhu/DHu.h
+++ b/include/manif/impl/dhu/DHu.h
@@ -1,0 +1,227 @@
+#ifndef _MANIF_MANIF_DHU_H_
+#define _MANIF_MANIF_DHU_H_
+
+#include "manif/impl/dhu/DHu_base.h"
+
+namespace manif {
+
+// Forward declare for type traits specialization
+
+template <typename _Scalar> struct DHu;
+template <typename _Scalar> struct DHuTangent;
+
+namespace internal {
+
+//! Traits specialization
+template <typename _Scalar>
+struct traits<DHu<_Scalar>>
+{
+  using Scalar = _Scalar;
+
+  using LieGroup = DHu<_Scalar>;
+  using Tangent  = DHuTangent<_Scalar>;
+
+  using Base = DHuBase<DHu<_Scalar>>;
+
+  static constexpr int Dim = LieGroupProperties<Base>::Dim;
+  static constexpr int DoF = LieGroupProperties<Base>::DoF;
+  static constexpr int RepSize = 8;
+
+  using DataType = Eigen::Matrix<Scalar, RepSize, 1>;
+
+  using Jacobian       = Eigen::Matrix<Scalar, DoF, DoF>;
+  using Transformation = Eigen::Matrix<Scalar, 4, 4>;
+  using Rotation       = Eigen::Matrix<Scalar, Dim, Dim>;
+  using Translation    = Eigen::Matrix<Scalar, Dim, 1>;
+  using Vector         = Eigen::Matrix<Scalar, Dim, 1>;
+};
+
+} // namespace internal
+} // namespace manif
+
+namespace manif {
+
+//
+// LieGroup
+//
+
+/**
+ * @brief Represent an element of DHu.
+ */
+template <typename _Scalar>
+struct DHu : DHuBase<DHu<_Scalar>>
+{
+private:
+
+  using Base = DHuBase<DHu<_Scalar>>;
+  using Type = DHu<_Scalar>;
+
+protected:
+
+  using Base::derived;
+
+public:
+
+  MANIF_MAKE_ALIGNED_OPERATOR_NEW_COND
+
+  MANIF_COMPLETE_GROUP_TYPEDEF
+  using Translation = typename Base::Translation;
+  using Quaternion = Eigen::Quaternion<Scalar>;
+  using Real = Quaternion;
+  using Dual = Quaternion;
+
+  MANIF_INHERIT_GROUP_API
+  using Base::transform;
+  using Base::rotation;
+  using Base::normalize;
+
+  DHu()  = default;
+  ~DHu() = default;
+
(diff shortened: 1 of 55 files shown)


artivis:
Draft implementation (class layout and a few functions in b9f046007476b3ca72679445c942d2aca99f9e9c).

joansola:
We should start by `exp`. Did you find any suitable source of info? I don't have a very sharp mind lately and would prefer to copy-paste as much as possible from reputable sources, rather than develop from scratch (which is much more fun, funnily).

I guess after `expe we'll be able to develop the `adj`oint. This is normally easy. 

The trouble will be as always with `J_l` and `J_r` ....... I fear these.

Jacobians are easy once we have the blocks above. We can try to simplify the expressions resulting from the chain rule. Often we are luck with these simplifications.

joansola:
Nice job BTW! how do you code so quickly? Did you create a template or script for new classes in `manif`?

include/manif/impl/dhu/DHu.h
```diff
@@ -0,0 +1,227 @@
...
+   * @param[in] q A unit quaternion.
+   * @throws manif::invalid_argument on un-normalized complex number.
+   */
+  DHu(const Real& real, const Dual& dual);
+
+  /**
+   * @brief Constructor given a translation and a unit quaternion.
+   * @param[in] t A translation vector.
+   * @param[in] q A unit quaternion.
+   * @throws manif::invalid_argument on un-normalized complex number.
```
joansola, line 100: un-normalized *complex number --> un-normalized quaternion

include/manif/impl/dhu/DHu_base.h
```diff
@@ -0,0 +1,411 @@
...
+}
+
+template <typename _Derived>
+void DHuBase<_Derived>::normalize()
+{
+  const Scalar n = coeffs().template head<4>().norm();
+  coeffs() /= n;
+  // real and dual parts are orthogonal
+  coeffs().template tail<4>() -= (coeffs().template head<4>().dot(coeffs().template tail<4>())) * coeffs().template head<4>();
+}
```
joansola, line 323: I did not understand this part in the paper. What are we to normalize? The rotational quaternion (4 components) and then the translational quaternion by the same scaling (4 more components)? Or all the real+dual quaterninon (8 components)? Or both the real and dual normalized separately?

That is, at the end of normalize(), what should we have:

1) | Real | = 1 and | Dual | = anything (just as in the code)
2) | (Real,Dual) | = 1
3) | Real | = 1 AND | Dual | = 1

??

artivis:
> We should start by `exp`. Did you find any suitable source of info? I don't have a very sharp mind lately and would prefer to copy-paste as much as possible from reputable sources, rather than develop from scratch (which is much more fun, funnily).

No I haven't found yet a source for `exp`/`log`, I'll look for it specifically. Definitely agree on developing as little as possible from scratch. So far I relied on [this source](http://web.cs.iastate.edu/~cs577/handouts/dual-quaternion.pdf).

> how do you code so quickly? Did you create a template or script for new classes in manif?

Ahah that's one of the many advantages of `manif`. Between the code standardization and the paper formalism more than half the job is done. I do not have a script, I simply copy an existing class (e.g. here `SE3` and related) and do some renaming accordingly. I could make a script but I'm not sure it is worth it time-wise.

include/manif/impl/dhu/DHu_base.h
artivis: In [this source](http://web.cs.iastate.edu/~cs577/handouts/dual-quaternion.pdf) the author defines 3 conjugates (Sec.5) and more specifically the second conjugate that "follows from the classical quaternion conjugation": `σ∗ = p∗ + ϵq∗`. Sec.6 states that "a dual quaternion is unit if σ ⊗ σ∗ = 1" which implies that the real part `p` must be a unit quaternion and the dual part `p` must be orthogonal to the real part `p`. So at the end of `normalize()` we should have:

4.a |Real| = 1
4.b Real . Dual = 0

These are the conditions I'm checking for at construction (see [here](https://github.com/artivis/manif/pull/165/commits/b9f046007476b3ca72679445c942d2aca99f9e9c#diff-f72abf273a625bdfe02f36e08d37c8daR396)). And I'm testing `normalize()` in [this test](https://github.com/artivis/manif/pull/165/commits/b9f046007476b3ca72679445c942d2aca99f9e9c#diff-a2ad8132fa38f63df0d82a193d4e92b7R132) which seems to pass.

joansola: How is the product in 4b defined? Is it regular quaternion product?

artivis: Sec.6 Eq.21 of the aforementioned reference: `p0q0 + p1q1 + p2q2 + p3q3 = 0`

joansola: OK I see, so they are orthogonal as 4-vectors: p.tr * q = 0

joansola:
exp can be found in 5.1 [here](https://openresearch.lsbu.ac.uk/download/76d2f90e65982a5246f866b96fed1f75b3a81a63dff996c182327a39e6b01376/231115/seligRev1.pdf)

joansola:
`log` in 8.6

PR #167, opened 2020-09-02
Update appveyor.yml

GiulioRomualdi:
Fix the appveyor compilation by changing the mirror used to get libeigen. Please check here for further information: http://eigen.tuxfamily.org/index.php?title=News:Migration_to_GitLab.com_scheduled_on_the_December_4th

PR #169, opened 2020-09-02
Bugfix in SE3TangentBase::asSO3() function

### LINKED ISSUE #168: Bug in SE3TangentBase::asSO3() function
The following code 
https://github.com/artivis/manif/blob/b1c82befd066fb4e954742bab56c7795168bdaab/include/manif/impl/se3/SE3Tangent_base.h#L123-L126
should be modified as 
```cpp
 Eigen::Map<SO3Tangent<Scalar>> asSO3() 
 { 
   return Eigen::Map<SO3Tangent<Scalar>>(coeffs().data()+3); 
 } 
```

GiulioRomualdi:
This PR fixes #168. 

Note on windows is failing because of #167

### DIFF
--- a/include/manif/impl/se3/SE3Tangent_base.h
+++ b/include/manif/impl/se3/SE3Tangent_base.h
@@ -122,7 +122,7 @@ public: /// @todo make protected
 
   Eigen::Map<SO3Tangent<Scalar>> asSO3()
   {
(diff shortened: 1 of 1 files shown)


artivis:
Good catch. This highlights that our test suite still has room for improvement!

PR #173, opened 2020-09-13
remove user defined destructors

mindbeast:


mindbeast:
Compiles locally fine for me. Perhaps this is because trusty has an older stdlib?

artivis:
Closing in favor of #147.

PR #183, opened 2020-12-08
fix compilation in release

### LINKED ISSUE #181: Problems building manif with examples in release mode
Hi there,

I receive compiler errors when following these steps on the latest version of the `devel` branch:
```
cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_EXAMPLES=ON ..
make
```

The errors are of the form (similar errors appear for `SE2`, `SE3`, etc.):
```
In file included from /path/to/manif/repo/include/manif/impl/so3/SO3.h:4:0,
                 from /path/to/manif/repo/include/manif/impl/so3/SO3_map.h:4,
                 from /path/to/manif/repo/include/manif/impl/se3/SE3_base.h:6,
                 from /path/to/manif/repo/include/manif/SE3.h:9,
                 from /path/to/manif/repo/examples/se3_sam_selfcalib.cpp:183:
/path/to/manif/repo/include/manif/impl/so3/SO3_base.h: In static member function ‘static void manif::internal::AssignmentEvaluatorImpl<manif::SO3Base<_Derived> >::run_impl(const T&)’:
/path/to/manif/repo/include/manif/impl/so3/SO3_base.h:402:53: error: typedef ‘using Scalar = typename manif::SO3Base<_Derived>::Scalar’ locally defined but not used [-Werror=unused-local-typedefs]
     using Scalar = typename SO3Base<Derived>::Scalar;
```

artivis:
closes #181

### DIFF
--- a/include/manif/impl/se2/SE2_base.h
+++ b/include/manif/impl/se2/SE2_base.h
@@ -414,11 +414,12 @@ struct AssignmentEvaluatorImpl<SE2Base<Derived>>
   static void run_impl(const T& data)
   {
     using std::abs;
-    using Scalar = typename SE2Base<Derived>::Scalar;
-    MANIF_ASSERT(abs(data.template tail<2>().norm()-Scalar(1)) <
-                 Constants<Scalar>::eps_s,
-                 "SE2 assigned data not normalized !",
-                 invalid_argument);
+    MANIF_ASSERT(
+      abs(data.template tail<2>().norm()-typename SE2Base<Derived>::Scalar(1)) <
+      Constants<typename SE2Base<Derived>::Scalar>::eps_s,
+      "SE2 assigned data not normalized !",
+      invalid_argument
+    );
   }
 };
 
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # manif
 ## A small header-only library for Lie theory.
 
-[![travis](https://travis-ci.com/artivis/manif.svg?branch=devel)](https://travis-ci.com/artivis/manif)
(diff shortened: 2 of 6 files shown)


artivis:
@joansola This is ready. Nevermind the pending test it is not going to run but passed previously.

include/manif/impl/se2/SE2_base.h
```diff
@@ -414,11 +414,12 @@ struct AssignmentEvaluatorImpl<SE2Base<Derived>>
...
-    MANIF_ASSERT(abs(data.template tail<2>().norm()-Scalar(1)) <
-                 Constants<Scalar>::eps_s,
-                 "SE2 assigned data not normalized !",
-                 invalid_argument);
+    MANIF_ASSERT(
+      abs(data.template tail<2>().norm()-typename SE2Base<Derived>::Scalar(1)) <
+      Constants<typename SE2Base<Derived>::Scalar>::eps_s,
+      "SE2 assigned data not normalized !",
+      invalid_argument
+    );
```
joansola, line 422: I guess the RELEASE mode was playing tricks here regarding the using directive?

joansola (approved):
I see the same patch is applied in a number of places. I dont fully understand the problem nor the solution but it seems that the optimizations in RELEASE mode were playing tricks with the code. I approve.

include/manif/impl/se2/SE2_base.h
artivis: The `MANIF_ASSERT` macro, in release, resolve to a blank (`#def MANIF_ASSERT() ((void)0)`). Because of that the local `Scalar` type was defined but not used, which was picked up as an error by the strict compilation flags used in our tests.

PR #186, opened 2020-12-10
Fix missing return in const data()

artivis:
An old stupid bug spotted there [PVS-Studio - "Checking a Header-Only C++ Library Collection (awesome-hpp)"](https://www.viva64.com/en/b/0770/#IDB5EB09C8DA).


artivis:
@joansola This is ready for merge.

joansola (approved):
OK

PR #188, opened 2020-12-19
Fix LieGroupBase::act signature

artivis:
Fix the signature of `LieGroupBase::act`.

### DIFF
--- a/include/manif/impl/lie_group_base.h
+++ b/include/manif/impl/lie_group_base.h
@@ -158,9 +158,10 @@ public:
    * @param[out] -optional- J_vout_v Jacobian of the new object wrt input object.
    * @return
    */
-  Vector act(const Vector& v,
-             OptJacobianRef J_vout_m = {},
-             OptJacobianRef J_vout_v = {}) const;
+  template <typename _EigenDerived>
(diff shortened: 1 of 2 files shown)


include/manif/impl/lie_group_base.h
```diff
@@ -601,10 +602,13 @@ LieGroupBase<_Derived>::adj() const
...
-LieGroupBase<_Derived>::act(const Vector& v,
-                            OptJacobianRef J_vout_m,
-                            OptJacobianRef J_vout_v) const
+LieGroupBase<_Derived>::act(
+  const Eigen::MatrixBase<_EigenDerived>& v,
+  tl::optional<Eigen::Ref<Eigen::Matrix<Scalar, Dim, DoF>>> J_vout_m,
+  tl::optional<Eigen::Ref<Eigen::Matrix<Scalar, Dim, Dim>>> J_vout_v
+) const
 {
   return derived().act(v, J_vout_m, J_vout_v);
```
joansola, line 613: what about the signature in derived classes?

artivis: There are correct and unit tested. The issue here really is when calling `act` from a `LieGroup<Derived>` object. This use-case is not unit tested at the moment but I'll look into that.

PR #179, opened 2020-12-07
[WIP] Add Python bindings to manif

artivis:
# manifpy 

Add Python bindings to manif.

Install,

```terminal
sudo apt install -y libeigen3-dev cmake build-essential python3-pip
python3 -m pip install --upgrade pip
python3 -m pip install pytest "pybind11[global]"
cd manif/
python3 -m pip install -r requirements.txt
python3 -m pip install .
```

Test,

```
cd manif/
pytest
```

Use,

```python
from manifpy import SE3

state = SE3.Identity()
...
```

### `rjacinv/ljacinv`

The functions `rjacinv/ljacinv` are not exposed at the moment. They are too complicated to wrap for little benefit.
This will be fixed in the future after the internal API redesign (see e.g. #150).

### `Eigen/Geometry`-related API

The C++ API related to `Eigen/Geometry` is not exposed. The rational is that this API only exists as a convenience to interface with existing classes in `Eigen` (`Quaternion`/`Transform` etc).
The `Eigen` counterpart in Python is `numpy` which does not have such classes. We could wrap the `Eigen` objects to expose them to Python but that does not have much interest since they wont play nice with `numpy` unless we put some extra effort.
At the moment this seems to be too much effort for little outcome, especially since the different examples (`_localization`/`_sam`) were ported to Python without these classes and related API.

For this reason this will be left to future work or contribution.

## Todo list

Bindings:
- [x] manif group-base API
- [x] manif tangent-base API
  - missing `rjacinv/ljacinv`
- [x] group specific API 
  - missing `Eigen/Geometry`-related
- [x] port common test suite to python
- [ ] port group specific tests to python

Misc:
- [x] python packaging
- [ ] documentation
  - [x] Python doc generation
  - [x] unified manif website
  - [ ] automatic doc update in CI
- [x] examples
  - [x] se2_localization
  - [x] se3_localization
  - [x] se2_sam
  - [x] se3_sam
  - [x] se3_sam_selfcalib
  - [x] se2_3_localization

Blocked by #190
Closes #187

PR #190, opened 2020-12-20
Disambiguate v/w/a -> lin/ang/lin2

artivis:
Disambiguate

- `SO3/SE3/SE_2_3::v` rename it `::lin`
- `SE3/SE_2_3::w` rename it `::ang`
- `SE_2_3::a` rename it `::lin2`

as discussed [here](https://github.com/artivis/manif/pull/179#issuecomment-748513087) and in this PR.

### DIFF
--- a/include/manif/impl/se3/SE3Tangent_base.h
+++ b/include/manif/impl/se3/SE3Tangent_base.h
@@ -28,10 +28,10 @@ public:
   MANIF_TANGENT_TYPEDEF
   MANIF_INHERIT_TANGENT_OPERATOR
 
-  using LinVel = typename DataType::template FixedSegmentReturnType<3>::Type;
-  using AngVel = typename DataType::template FixedSegmentReturnType<3>::Type;
-  using ConstLinVel = typename DataType::template ConstFixedSegmentReturnType<3>::Type;
-  using ConstAngVel = typename DataType::template ConstFixedSegmentReturnType<3>::Type;
+  using LinBlock = typename DataType::template FixedSegmentReturnType<3>::Type;
+  using AngBlock = typename DataType::template FixedSegmentReturnType<3>::Type;
+  using ConstLinBlock = typename DataType::template ConstFixedSegmentReturnType<3>::Type;
+  using ConstAngBlock = typename DataType::template ConstFixedSegmentReturnType<3>::Type;
 
   using Base::data;
   using Base::coeffs;
@@ -105,12 +105,12 @@ public:
   // SE3Tangent specific API
 
   //! @brief Get the linear part.
-  LinVel linVel();
-  const ConstLinVel linVel() const;
+  LinBlock lin();
+  const ConstLinBlock lin() const;
 
   //! @brief Get the angular part.
-  AngVel angVel();
-  const ConstAngVel angVel() const;
+  AngBlock ang();
+  const ConstAngBlock ang() const;
 
 //  Scalar x() const;
 //  Scalar y() const;
@@ -153,7 +153,7 @@ SE3TangentBase<_Derived>::exp(OptJacobianRef J_m_t) const
   }
 
   /// @note Eq. 10.93
-  return LieGroup(asSO3().ljac()*linVel(), asSO3().exp().quat());
+  return LieGroup(asSO3().ljac()*lin(), asSO3().exp().quat());
 }
 
 template <typename _Derived>
@@ -310,8 +310,8 @@ SE3TangentBase<_Derived>::smallAdj() const
   /// considering vee(log(g)) = (v;w)
 
   Jacobian smallAdj;
-  smallAdj.template topRightCorner<3,3>() = skew(linVel());
-  smallAdj.template topLeftCorner<3,3>() = skew(angVel());
+  smallAdj.template topRightCorner<3,3>() = skew(lin());
+  smallAdj.template topLeftCorner<3,3>() = skew(ang());
   smallAdj.template bottomRightCorner<3,3>() = smallAdj.template topLeftCorner<3,3>();
   smallAdj.template bottomLeftCorner<3,3>().setZero();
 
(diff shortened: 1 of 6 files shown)


prashanthr05:
@artivis I just had a suggestion as a `manif` user.

Do you think it would be better to use a deprecation warning for the v/w/a methods in the current release and remove them in a new release, so that the `manif` users might cleanly transition to the new methods? Or do you think it would block the ambiguity fix against the innerWeightedNorm ->w() and quaternion -> w() ?

artivis:
Hi @prashanthr05,

I would normally agree with you and plan for a smooth deprecation but my feeling is that the issue at hand is sufficiently major to break API abruptly. This is all on me for not foreseeing this earlier.
With the current state, calling `tangent.w()` may result in a compilation error (I'd say in the best case) or it may results in calling two different functions depending on the tangent object - `se2_tangent.w()` returns the inner weight matrix while `se_2_3_tangent.w()` returns the angular velocity block. Note that calling the inner weight matrix of `se_2_3_tangent` is therefore hidden.

What I can do is issue a new tag, `0.0.3`, before merging those upcoming breaks so that users can at least refer to it.

prashanthr05:
Ah, I see.  I had not realized it was rather actually a critical bug fix than just a naming clearance in the API.  I had not thought about the inner weight matrix of `se_2_3_tangent` as well before.
Thank you for the details! It gives me some clarity.

> What I can do is issue a new tag, 0.0.3, before merging those upcoming breaks so that users can at least refer to it.

I think this would be helpful. Thanks!

joansola:
I have a comment regarding the names for the tangent elements in terms of velocities.

Let us denote `X` a group element, `v` a velocity element, and `dx = v*dt` a tangent displacement of `X`.

Clearly, velocities `v` are in the tangent space, this is by definition.

But, we actually have `X = exp(v*dt)` and therefore `v*dt = log(X)`.

It is usually the case that in the tangent space we store values of `v*dt = dx`, and not velocity values.

In this regard, having non-abstract names for the tangent space is misleading. For example, let us have:

```c++
SE3        X = SE3::Random();
SE3Tangent a = X.log();
Vector3    d = a.linVel();
```

In this case, `d` is a displacement, but we have to retrieve it using `linVel()`, which is odd.

Likewise, we can have the code:

```c++
SE3Tangent a;
a.linVel() = v;
a.angVel() = w;
SE3 X = (a*dt).exp();
```
which makes complete sense in terms of velocities.

Suggestions welcome.

joansola:
To address the comment above, I suggest the following

```c++
SE3Tangent dx, v;
displacement = dx.lin();
theta        = dx.ang();
lin_vel      = v.lin();
ang_vel      = v.ang();
```

In case of

joansola:
We cal also use `v_dt()`, `w_dt()` and `a_dt()` or similar arrangements, but again this is not abstract enough and we cannot deal with all cases (see second example in original comment).

artivis:
You are right that the current proposal is a terrible wording. And since we are trying to disambiguate `v`/`w`, I don't think `v_dot`/`w_dot` are good either.
How about
```cpp
SE3_2Tangent tan;
lin_block      = tan.linBlock();
ang_block      = tan.angBlock();
lin_block      = tan.accBlock();
```
This has the advantage of being abstract and programatically meaningful.

joansola:
`lin` and `ang` are OK, but `acc` is not. See that the term `acc` in case of a `log()` has units `a*dt` that is `m/s`. It is a `velocity_increment`, the same way as `lin` is a `position increment` and `ang` is `rotation_increment`.

Besides, the `Block` addition is redundant, but if it adds clarity is OK.

joansola:
actually, in SE3_2, the group element is `X = [ R T1 T2 ; 0 1 0 ; 0 0 1]`, and T1 and T2 do not have a special meaning. Therefore, we could use `lin` and `lin_2`, in the same way the group is called `SE3_2`.

artivis:
Alright so we have 2 out of 3. 
`tan.lin_2Block()` would be very weird and we're not using underscore in our function naming so that would be `tan.lin2Block()` but I don't like it much either...

joansola:
What about `lin()`, `ang()`, `lin2()` ? 

Adding the `Block` makes `lin2Block` seem like a translation function from `lin` type to `block` type, which makes no sense.

joansola:
`lin2`, `linAlt`, `linTwo`, `linOther`

prashanthr05:
If I may add an opinion, I suppose `lin2()` looks descriptive enough if we document the lines of @joansola 

physically,
> It is a `velocity_increment`, the same way as `lin` is a `position increment` and `ang` is `rotation_increment`.

mathematically (abstract),
> actually, in SE3_2, the group element is `X = [ R T1 T2 ; 0 1 0 ; 0 0 1]`, and T1 and T2 do not have a special meaning. Therefore, we could use `lin` and `lin_2`, in the same way the group is called `SE3_2`.

artivis:
> What about lin(), ang(), lin2() ?

What are the associated types then? `Lin`, `Ang`, `Lin2`? `LinBlock`, `AngBlock`, `Lin2Block`? :s

joansola:
Types? They are all vectors, do we need to give them a special type?

artivis:
> Types? They are all vectors, do we need to give them a special type?

I don't mean an actual types but a type aliases like [those](https://github.com/artivis/manif/blob/f62ddaf99d71f9bdc4fb133be4b1c1f51ffccfae/include/manif/impl/se3/SE3Tangent_base.h#L31). That facilitate declaring variables should one need that.

artivis:
@prashanthr05 FYI I released tag [0.0.3](https://github.com/artivis/manif/releases/tag/0.0.3), so I'll start merging API-breaking PRs.

joansola:
Uhm really dont know. Declaring a type as `Lin`, `Ang`, `LinBlock` seems kind of confusing to me.

What about `LinVec`, `AngVec`, `Lin2Vec` ? Or even, use type `LinVec` also for `lin2()`, which makes a lot of sense because it's the same type.

artivis:
If I'm not mistaken, Eigen return a type alias `VectorBlock`. That's why I lean toward `LinBlock` and `AngBlock`. 
I'm fine with using the same alias for both `lin` and `lin2` here.

joansola:
> If I'm not mistaken, Eigen return a type alias `VectorBlock`. That's why I lean toward `LinBlock` and `AngBlock`.
> I'm fine with using the same alias for both `lin` and `lin2` here.

OK for me then

prashanthr05:
> @prashanthr05 FYI I released tag [0.0.3](https://github.com/artivis/manif/releases/tag/0.0.3), so I'll start merging API-breaking PRs.

@artivis  Thank you for the heads up :)

joansola (approved):
All seem OK to me. Approving

PR #194, opened 2020-12-30
Minor fixes: use known inverses, fix ambiguous ternary types

pettni:
The ambiguous ternary type issue occurs when using a datatype that uses expression types to represent operations (in this case minus and arctan2), which is the case with many libraries for auto-differentiation.

### DIFF
--- a/include/manif/impl/se2/SE2_base.h
+++ b/include/manif/impl/se2/SE2_base.h
@@ -252,7 +252,7 @@ SE2Base<_Derived>::log(OptJacobianRef J_t_m) const
   if (J_t_m)
   {
     // Jr^-1
-    (*J_t_m) = tan.rjac().inverse();
+    (*J_t_m) = tan.rjacinv();
   }
 
   return tan;
--- a/include/manif/impl/se3/SE3_base.h
+++ b/include/manif/impl/se3/SE3_base.h
@@ -296,13 +296,13 @@ SE3Base<_Derived>::log(OptJacobianRef J_t_m) const
   const SO3Tangent<Scalar> so3tan = asSO3().log();
 
   Tangent tan((typename Tangent::DataType() <<
-               so3tan.ljac().inverse()*translation(),
+               so3tan.ljacinv()*translation(),
                so3tan.coeffs()).finished());
 
   if (J_t_m)
   {
     // Jr^-1
-    (*J_t_m) = tan.rjac().inverse();
+    (*J_t_m) = tan.rjacinv();
   }
 
   return tan;
--- a/include/manif/impl/se_2_3/SE_2_3_base.h
+++ b/include/manif/impl/se_2_3/SE_2_3_base.h
@@ -256,14 +256,14 @@ SE_2_3Base<_Derived>::log(OptJacobianRef J_t_m) const
   const SO3Tangent<Scalar> so3tan = asSO3().log();
 
   Tangent tan((typename Tangent::DataType() <<
-               so3tan.ljac().inverse()*translation(),
+               so3tan.ljacinv()*translation(),
                so3tan.coeffs(),
-               so3tan.ljac().inverse()*linearVelocity()).finished());
+               so3tan.ljacinv()*linearVelocity()).finished());
 
   if (J_t_m)
   {
     // Jr^-1
-    (*J_t_m) = tan.rjac().inverse();
+    (*J_t_m) = tan.rjacinv();
   }
 
   return tan;
--- a/include/manif/impl/so3/SO3_base.h
+++ b/include/manif/impl/so3/SO3_base.h
@@ -194,8 +194,8 @@ SO3Base<_Derived>::log(OptJacobianRef J_t_m) const
      *            = atan(-sin(angle), -cos(angle))
      */
     const Scalar two_angle = Scalar(2.0) * ((cos_angle < Scalar(0.0)) ?
-                                 atan2(-sin_angle, -cos_angle) :
-                                 atan2( sin_angle,  cos_angle));
+                                 Scalar(atan2(-sin_angle, -cos_angle)) :
+                                 Scalar(atan2( sin_angle,  cos_angle)));
 
     log_coeff = two_angle / sin_angle;
   }


artivis (approved):
lgtm

artivis:
@pettni Which autodiff lib raised this issue?

pettni:
> @pettni Which autodiff lib raised this issue?

I added one more commit after also testing SE2 (I had previously only looked at SE3). The commit message explains the issue.

In case you are curious about manif compatability with autodiff I'll provide some more detail. I have been looking at a couple of different libraries. With these three commits most seem compatible with manif (though my tests are not exhaustive).

* [Adept](https://github.com/rjhogan/Adept-2): Compatible
* [ADOL-C](https://github.com/coin-or/ADOL-C/): Compatible in tapeless mode (taped mode does not overload ```abs``` for some reason)
* [autodiff](https://github.com/autodiff/autodiff/): Compatible
* Ceres: Compatible
* [CppAD](https://github.com/coin-or/CppAD): Compatible but might need re-taping when conditionals change
* [CppADCodeGen](https://github.com/joaoleal/CppADCodeGen/): Had trouble with this one and it was throwing exceptions, I believe when calling trigonometry functions
* [Sacado](https://github.com/trilinos/Trilinos): Compatible

I'm not an expert on CppAD, but there are ways to define custom functions and conditional expressions that might circumvent the issues above. I might look into this later since CppADCodeGen seems much faster in most cases (after compiling the generated code once).

To make it work I had to overload the manif ```Constants``` trait like you have already done for the Ceres jet type, since the autodiff types are not literal and ```constexpr```-compatible. It would be convenient if manif came with two specializations of the ```Constants``` trait, one for literal and one for non-literal types.

However, two libraries (adept and adolc) use internal reference counting and it caused some issues to have their scalars defined in the trait. I instead just defined them as doubles, e.g.
```
struct Constants<adept::Active<_Scalar>>
{
  static constexpr double eps = 1e-10;
  static constexpr double eps_s = 1e-15;
};
```

include/manif/impl/eigen.h
```diff
@@ -134,9 +134,8 @@ struct traitscast<Eigen::Matrix<_Scalar, _Rows, _Cols, _Options, _MaxRows, _MaxC
  *             | x  0 |
  */
 template <typename _Scalar>
-typename std::enable_if<std::is_arithmetic<_Scalar>::value,
-                        Eigen::Matrix<_Scalar, 2, 2>>::type
-skew(const _Scalar v)
+Eigen::Matrix<_Scalar, 2, 2>
+skew2x2(const _Scalar v)
```
artivis, line 138: Can you please provide an example which raises the issue here? I'd prefer to find a better suited sfinae condition rather than renaming the function. In case such condition doesn't exists, or is too wobbly, I would rename the function `skew2` and the other one `skew3` to better stick with `Eigen` naming.

artivis:
Thanks for the detailed list of autodiff libraries. As you pointed, `manif` is not fully compatible with tape-based libraries (e.g. `CppAD`) at the moment. The reason is that conditional branching break the tape and special care must be taken. Long ago I have been working on getting `CppAD` full support (branch `cppad`, see [here](https://github.com/artivis/manif/compare/08537f66e0dc867b2981ee5b3a4679fb65636953...c532ac7b2be15266f7e32b0e045efb6101a88f68)) but never quite finish it. 

Would you be up to write a few unit tests for some of the libraries you listed? That would ensure the compatibility is not broken later down the road.

include/manif/impl/eigen.h
pettni: Here's an example:

```
#include <manif/impl/traits.h>
#include <manif/impl/eigen.h>

// example "autodiff" type
struct MyDouble
{
  MyDouble(double d)
    : d_(d)
  {}

  double d_;
};

int main()
{
  Eigen::Matrix<double, 6, 1> v = Eigen::Matrix<double, 6, 1>::Zero();

  // these should call the scalar version
  manif::skew(1.0);
  manif::skew(MyDouble(1));

  // these should call the vector version
  manif::skew(Eigen::Vector3d{1, 2, 3});
  manif::skew(v.head<3>());
}
```
It seems like the scalar version should be called for all types that are not Eigen types, but I couldn't find a good condition to express that something is not an Eigen type. I tried ```!is_base_of<MatrixBase<T>, T>``` but this does not work since ```VectorBlock``` does not inherit from ```MatrixBase```.

Let me know if you find a working condition, otherwise I'll go ahead with the renaming.

pettni:
I could for sure include some unit tests. How would you go about installing the corresponding autodiff library? Install in the CI job like how it's done with Ceres?

Also which AD tool would you include? autodiff is probably my favorite and header-only, but needs c++17. Adept, adol-c and sacado are a bit cumbersome to install, so maybe CppAD or just using the internal Ceres autodiff is a good option?

include/manif/impl/eigen.h
artivis: Thanks for the small example, I'll have a look at what's possible to do.  
At the end of the day we can always have a custom traits in `manif` that's specialized for each AD scalar type, something along the line:
```cpp
template <typename Scalar> struct is_ad : std::integral_constant<bool, false> { };
...
template <> struct is_ad<MyDouble> : std::integral_constant<bool, true> { };
...
template <typename Scalar>
typename std::enable_if<std::is_arithmetic<Scalar>::value || is_ad<Scalar>::value, Matrix>
skew( ... ) { ... }
```

artivis:
> How would you go about installing the corresponding autodiff library? Install in the CI job like how it's done with Ceres?

Yeah that's most likely the way. Simply make sure that, like `Ceres`, the dependencies you may introduce are 'soft'.

> Also which AD tool would you include?

I agree with you in that having `autodiff`, `CppAD` and `Ceres::Jet` tested would be awesome. Upon further looking at the `autodiff` library, it seems to offer both forward and reverse mode. Are both methods tape-based or only the reverse mode?
Note that `Ceres::Jet` is already [somewhat tested for SE2](https://github.com/artivis/manif/blob/devel/test/ceres/gtest_se2_autodiff.cpp). There is also a [tester class](https://github.com/artivis/manif/blob/devel/test/ceres/ceres_test_utils.h) which appears to be testing nothing (:facepalm:) which could serve as a basis for the other AD types test suite.

artivis:
After realizing that the `Ceres` test suite was not actually testing anything, I got started on implementing a few in #195. That may be a starting point to test other AD libraries.

artivis:
@pettni If you agree, could you please revert the last commit about the `skew` function and open a ticket about it? So that I'll merge the original small fixes asap.

pettni:
> @pettni If you agree, could you please revert the last commit about the `skew` function and open a ticket about it? So that I'll merge the original small fixes asap.

Sure, done

artivis (approved):

artivis:
Thanks for the fixes!

PR #195, opened 2021-01-07
Adding Ceres::Jet unit tests

artivis:
Using `Ceres::Jet` (autodiff), compute the Jacobians below and compare them against `manif` analytic one.

All Jacobians in `manif` are tested:
- inverse
- rplus
- lplus
- rminus
- lminus
- log
- compose
- between
- act

Depends on #199 for the `act` test.

### DIFF
--- a/include/manif/impl/rn/Rn_base.h
+++ b/include/manif/impl/rn/Rn_base.h
@@ -1,10 +1,8 @@
 #ifndef _MANIF_MANIF_RN_BASE_H_
 #define _MANIF_MANIF_RN_BASE_H_
 
-#include "manif/impl/so2/SO2_properties.h"
+#include "manif/impl/rn/Rn_properties.h"
 #include "manif/impl/lie_group_base.h"
-
-#include <iostream>
 
 namespace manif {
 
--- a/include/manif/impl/traits.h
+++ b/include/manif/impl/traits.h
@@ -48,6 +48,19 @@ struct traitscast<_Class<_Scalar>, _NewScalar>
   using cast = _Class<_NewScalar>;
 };
 
+/**
+ * @brief Specialization for Rn.
+ */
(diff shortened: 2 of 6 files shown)


artivis:
@joansola FYI.
Altho it seems to work, we will have to discuss (later) a jacobian in this PR. I'm computing a local-parameterization **inverse** Jacobian that I'm not sure about - or a local-deparameterization Jacobian of some sort.

joansola:
We DID discuss this Jacobian parametrization inverse no? Is there any other point of conflict?

joansola:
Where are the Jacobians tested / where are the results of these tests? Are there mere ctest cases?

artivis:
> Where are the Jacobians tested / where are the results of these tests? Are they mere ctest cases?

Yes they are ctests (a gtest class) wrapped in a macro for convenience, see [`ceres_test_utils.h`](https://github.com/artivis/manif/blob/97a65bb0585f471f5fc7dfc9a99cdac7456e13bb/test/ceres/ceres_test_utils.h).

artivis:
> We DID discuss this Jacobian parametrization inverse no? Is there any other point of conflict?

We did discuss it indeed and except for some notation we were agree in principle.

joansola:
What do you want me to review then? I can  just approve, if tests pass and the principles are agreed, I would approve. Your opinion?

artivis:
> What do you want me to review then? I can just approve, if tests pass and the principles are agreed, I would approve. Your opinion?

Yeah you can approve. Those tests were not really needed since we already are checking the Jacobians in `manif`. They test for the compatibility with `Ceres::Jet` more than the maths.

joansola:
Yea true. OK

joansola (approved):
All agreed

PR #199, opened 2021-01-08
Add traits `is_ad`

### LINKED ISSUE #198: No skew function available for custom scalar types
Custom scalar types don't necessarily specialize ```std::is_arithmetic``` (doing so in isolation seems to cause a host of other problems in some cases). This causes a problem with the SFINAE ```skew``` functions.

Example:
```
#include <manif/impl/traits.h>
#include <manif/impl/eigen.h>

// example "autodiff" type
struct MyDouble
{
  MyDouble(double d)
    : d_(d)
  {}

  double d_;
};

int main()
{
  Eigen::Matrix<double, 6, 1> v = Eigen::Matrix<double, 6, 1>::Zero();

  // these should call the scalar version
  manif::skew(1.0);
  manif::skew(MyDouble(1));

  // these should call the vector version
  manif::skew(Eigen::Vector3d{1, 2, 3});
  manif::skew(v.head<3>());
}
```

artivis:
Add a traits to detect when `Scalar` is an autodiff (AD) type.
Specialize it for `ceres::Jet`.

Fixes #198.

### DIFF
--- a/include/manif/ceres/ceres.h
+++ b/include/manif/ceres/ceres.h
@@ -10,6 +10,10 @@
 namespace manif {
 namespace internal {
 struct YOU_MUST_INCLUDE_MANIF_BEFORE_CERES_HELPER_HEADERS{};
+
+template <typename _Scalar, int _N>
+struct is_ad<ceres::Jet<_Scalar, _N>> : std::integral_constant<bool, true> { };
+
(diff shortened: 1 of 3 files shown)


artivis:
@joansola ready for approval.

PR #200, opened 2021-02-13
Miscellaneous small fixes

artivis:
Miscellaneous small fixes:

- group == operator constness
- tangent constexpr variables
- tl::optional missing return value
- examples: init random generator and reduce noise
- disable macos-11 in CI since the image is not ready yet (see https://github.com/actions/virtual-environments/issues/841)
- delete Travis config file since we're not using it anymore

PR #201, opened 2021-02-13
Add Python bindings

### LINKED ISSUE #187: Python Bindings discussion
@artivis 

I wanted to keep the PR more for review. So making a separate post.

Since I have some time over the Christmas break is there anything specifically you would like me to work on.  Just so I don't overlap with your work. From reading your PR I think we are missing the following.

- [ ] Create bindings for eigen classes. (Needed so SE3 can accept `Eigen::Quaternion` ...)
- [ ] SE3 specifics
- [ ] SE2 specifics
- [ ] Other classes specifics
- [ ] Python packaging
- [ ] Examples

The other two things are `numpy<-> eigen` compatability(beyond setting Numpy to `'F'`), and documentation. But I think we should hold off on those until the rest of the library is up and running. They also warrant some discussion as there are multiple ways to implement them.

artivis:
# manifpy 

Add Python bindings to manif.

Install,

```terminal
sudo apt install -y libeigen3-dev cmake build-essential python3-pip
python3 -m pip install --upgrade pip
python3 -m pip install pytest "pybind11[global]"
cd manif/
python3 -m pip install -r requirements.txt
python3 -m pip install .
```

Test,

```
cd manif/
pytest
```

Use,

```python
from manifpy import SE3

state = SE3.Identity()
...
```

### `rjacinv/ljacinv`

The functions `rjacinv/ljacinv` are not exposed at the moment. They are too complicated to wrap for little benefit.
This will be fixed in the future after the internal API redesign (see e.g. #150).

### `Eigen/Geometry`-related API

The C++ API related to `Eigen/Geometry` is not exposed. The rational is that this API only exists as a convenience to interface with existing classes in `Eigen` (`Quaternion`/`Transform` etc).
The `Eigen` counterpart in Python is `numpy` which does not have such classes. We could wrap the `Eigen` objects to expose them to Python but that does not have much interest since they wont play nice with `numpy` unless we put some extra effort.
At the moment this seems to be too much for the little outcome, especially since the different examples (`*_localization` / `*_sam`) were ported without issues.

For this reason this will be left to future work or contribution.

## Todo list

Bindings:
- [x] manif group-base API
- [x] manif tangent-base API
  - missing `rjacinv/ljacinv`
- [x] group specific API 
  - missing `Eigen/Geometry`-related
- [x] port common test suite to python
- [ ] port group specific tests to python (partially done)

Misc:
- [x] python packaging
- [x] documentation
  - [x] Python doc generation
  - [x] unified manif website
  - [x] automatic doc update in CI
- [x] examples
  - [x] se2_localization
  - [x] se3_localization
  - [x] se2_sam
  - [x] se3_sam
  - [x] se3_sam_selfcalib
  - [x] se2_3_localization

Cleanup version of #179

Comes after #200
Closes #187


artivis:
See #179 for discussion backlogs.

joansola (commented):
Working on MacOSX.

Great work!

How can I execute the python examples? I´d like to see them running before approving

artivis:
> How can I execute the python examples? I´d like to see them running before approving

First you need to install `manifpy` is not already done (see the PR description), then you can execute them as plain Python script, e.g. 
```bash
cd manif/examples
python3 se3_localization.py
```

joansola (approved):
Examples checked and working. Approving with congrats

artivis:
Post-approval update:
- rebased on `devel`
- update c++ doc to include freshly merged UKF-M example

PR #214, opened 2021-03-16
Set project version

artivis:
- bump CMake version
- set project version properly (fix CMake warning)
- setuptools project version fetched from package.xml


artivis:
@joansola This is ready for review/merge.

package.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package format="2">
   <name>manif</name>
-  <version>0.0.0</version>
+  <version>0.0.3</version>
```
joansola, line 4: Is it OK to require the version specification twice? Once in CMakeLists and once here?

artivis: Note that it is also specified in the `setup.py` altho there I could manage to retrieve the version from the `CMakeLists.txt`. It is indeed unfortunate to specify it in several places but I couldn't find a workaround. Note that they are used for the same purpose by two different tools.
`CMake` requires the version in the `CMakeLists.txt` when invoking e.g. `find_package(manif 0.0.3)` and similarly ROS requires it in the `package.xml` for the same versioning purpose

joansola: yeah ok I see. No problem. Approving

joansola (approved):

PR #217, opened 2021-03-20
Fixes for float support

### LINKED ISSUE #171: EPS is not suitable for float32
https://github.com/artivis/manif/blob/b2f777967f343b51ba954fa641c9a256564f536a/include/manif/constants.h#L51

hi, guys, i found this line will throw out an error because the quaternion not normalized, but i have used the quaternion normalize function provided with eigen. i think this problem maybe 1e-15 is not suitable for float32 value, 1e-6 will be more suitable for it.

### LINKED ISSUE #216: Support for float type
I am using manif for some computations in an existing project which is entirely working with fp32 numbers. However when I use they float type with any operations in manif, I more or less randomly (but very frequently) get the exception

```
terminate called after throwing an instance of 'manif::invalid_argument'
what():  SO3 assigned data not normalized !
```
This exception stems from the hardcoded eps for the double precision type in constants.h. Is it at all possible for you to support other types than double (e.g. by using std::numeric_limits<_Scalar>::epsilon())?

Best regards and thanks for the great library.

artivis:
- specialize `Constants` for float
- add `randQuat` and use it in SO3/SE3/SE_2_3 rather than having duplicated code. It calls `Eigen::Quaternion::UnitRandom` when available
- Test SO3 for float and relax some test tolerances

Closes #171
Closes #216


artivis:
@joansola This is ready for review/merge.

include/manif/impl/eigen.h
```diff
@@ -193,6 +193,34 @@ Eigen::Matrix<Scalar, 3, 1> randPointInBall(Scalar radius)
   );
 }
 
+template <typename Scalar>
+Eigen::Quaternion<Scalar> randQuat()
+{
+#if EIGEN_VERSION_AT_LEAST(3,3,0)
+
+  return Eigen::Quaternion<Scalar>::UnitRandom();
```
joansola, line 201: Long ago we discussed quat randomness. I seem to remember that Eigen implementation was perhaps non-uniform on the unit sphere. If I am right... does UnitRandom fix this?

artivis: The implementation we used until now (which is still available at the lines below) is a literal copy/paste from `Eigen::Quaternion<Scalar>::UnitRandom`. So this technically does not change anything.

joansola (approved):

PR #219, opened 2021-03-29
Install gtest helpers

artivis:
Rename & install gtest helpers (Eigen / manif).

It allows projects using `manif` to make use of the gtest helper macros such as `EXPECT_MANIF_NEAR( x, y )`.
They are not included by default with the global `manif.h` header file, one has to explicitly include them, e.g.

```cpp
#include <manif/gtest/gtest_manif_utils.h>
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,8 +147,9 @@ write_basic_package_version_file(
 
 # Config
 install(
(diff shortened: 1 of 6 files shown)


artivis:
@joansola hey, can I get an approval to merge this?

joansola (approved):
ok

PR #224, opened 2021-04-24
Run tests at/around identity

### LINKED ISSUE #223: ljacinv
https://github.com/artivis/manif/blob/devel/include/manif/impl/so3/SO3Tangent_base.h#L214
accroding to:
image

  if (theta_sq <= Constants<Scalar>::eps_s)
    return Jacobian::Identity() + Scalar(0.5) * W; 

Should be changed to

  if (theta_sq <= Constants<Scalar>::eps_s)
    return Jacobian::Identity() -  Scalar(0.5) * W; 
 
Is that right?

artivis:
The test suit is currently run on values set and random (`LieGroup::Random()`). This PR extend the test suit to also run the tests at identity (`LieGroup::Identity()`) and very close to it (`(Tangent::Random()*1e-8).exp()`) in order to properly test the small angles approximations paths. Doing so made the issue in #223 clearly appear. While at it, the SO3 left-Jac small angle approx was wrong too...

Closes #223

### DIFF
--- a/include/manif/algorithms/average.h
+++ b/include/manif/algorithms/average.h
@@ -53,7 +53,7 @@ template <template <typename LieGroup, typename...Args> class Container,
 LieGroup
 average_biinvariant(const Container<LieGroup, Args...>& points,
                     typename LieGroup::Scalar eps =
-                      Constants<typename LieGroup::Scalar>::eps_s,
+                      Constants<typename LieGroup::Scalar>::eps,
                     int max_iterations = 20)
 {
   using Scalar  = typename LieGroup::Scalar;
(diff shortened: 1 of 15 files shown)


joansola:
what do you mean by "SO3 left-Jac small-angle approx was wrong too"? Which other one was wrong? This PR only fixes SO3::ljac()...

artivis:
> what do you mean by "SO3 left-Jac small-angle approx was wrong too"? Which other one was wrong? This PR only fixes SO3::ljac()...

This fixes `SO3::ljac()` and `SO3::ljacinv()`, see https://github.com/artivis/manif/pull/224/commits/e05e446bb4e0d7e691a4b3a353cd8f02e261fa37

joansola (approved):
all good!

PR #105, opened 2019-09-19
faster implementation of SO3 random

dllu:
Following my [comment on sophus](https://github.com/strasdat/Sophus/issues/192#issuecomment-533065151) I suspected that [Marsaglia's rejection sampling](http://mathworld.wolfram.com/HyperspherePointPicking.html) is faster than the more elegant method that is currently being used.

So I implemented this method.

Below are some benchmarks. Please check out my branch and test on your machine and see if your experience matches mine. It seems we can get about 20% more performance.

I am confused why the same code for generating a random quaternion is duplicated in `SE3_base.h` and in `SO3_base.h`. If anyone has ideas for refactoring it so that it is no longer duplicated, I will be happy to hear.

## Benchmarks:

Benchmark code:

```c++
#include <Eigen/Dense>
#include <chrono>
#include <fstream>
#include <iostream>
#include <vector>

#include "manif/SO3.h"

int main() {
    std::ofstream fout("so3_randomization.txt");
    const auto tic = std::chrono::steady_clock::now();
    const size_t n = 20'000'000;
    std::vector<manif::SO3d, Eigen::aligned_allocator<manif::SO3d>> stuff;
    stuff.reserve(n);
    for (size_t i = 0; i < n; i++) {
        stuff.push_back(manif::SO3d::Random());
    }
    const auto toc = std::chrono::steady_clock::now();
    std::cout << "Time elapsed: "
              << std::chrono::duration<double>(toc - tic).count() << " s"
              << std::endl;

    for (const auto &zxcv : stuff) {
        fout << zxcv.w() << ", " << zxcv.x() << ", " << zxcv.y() << ", "
             << zxcv.z() << std::endl;
    }
    return 0;
}

```

| compiler | manif branch | result (seconds) |
| - | - | - |
| `clang++ -O3 -I../manif/include -I/usr/include/eigen3 -I../manif/external/lt -o test test.cpp` | devel | 1.62653 |
| `clang++ -O3 -I../manif/include -I/usr/include/eigen3 -I../manif/external/lt -o test test.cpp` | dllu/faster-so3-random | 1.0261 |
| `g++ -O3 -I../manif/include -I/usr/include/eigen3 -I../manif/external/lt -o test test.cpp` | devel | 1.46156 |
| `g++ -O3 -I../manif/include -I/usr/include/eigen3 -I../manif/external/lt -o test test.cpp` | dllu/faster-so3-random | 1.22057 |

## Compilers used:

```
manif-test » clang++ --version
[...]

PR #227, opened 2021-04-30
Add benchmarking & some opti

artivis:
- Add benchmark
- some small opti
- `SE2Tangent::rjacinv/ljacinv`
- `SE_2_3Tangent::rjacinv/ljacinv`
- fix average unit test
- run tests in debug and release on ubuntu.

This PR adds some micro benchmarks using [Google benchmark](https://github.com/google/benchmark). This allows to have some time-based benchmarks on a per-function basis so that one can compare different implementations.

The PR also contains numerous small optimizations, most of which do not actually show any benefit in the aforementioned benchmark. Those which do are listed below. 
Most noticeably, implementing `SE2Tangent::rjacinv/ljacinv` (~6x / ~4.3x) and `SE_2_3Tangent::rjacinv/ljacinv` (~4.5x / ~5x) are a big win.

It also remove some unnecessary cmake warning suppression for the test/examples, fix the `average` unit test that was disabled and finally it runs the unit test both in Debug and Release on Ubuntu.

Some numbers obtained through the benchmarking introduced in this PR:

| Group | Operation  |       | Before |       | After |
| :--- |   :---:   |   :---:   | :---: | :---: | :---: |
|SE2|rjacinv|||||
|(mean)|        | | 49.6 ns   | | 8.41 ns   |
|(median)|      | | 49.6 ns   | | 8.39 ns   |
|(stddev)|      | | 3.69 ns   | | 0.102 ns  |
||ljacinv|||||
||              | | 37.1 ns   | | 8.48 ns   |
||              | | 37.3 ns   | | 8.46 ns   |
||              | | 3.77 ns   | | 0.097 ns  |
|SE_2_3|rjacinv|||||
||              | | 1570 ns   | | 349 ns    |
||              | | 1539 ns   | | 354 ns    |
||              | | 72.8 ns   | | 9.76 ns   |
||ljacinv|||||
||              | | 1722 ns   | | 338 ns    |
||              | | 1706 ns   | | 337 ns    |
||              | | 68.4 ns   | | 11.0 ns   |
||adj|||||
||              | | 100 ns    | | 76.1 ns   |
||              | | 100 ns    | | 75.8 ns   |
||              | | 1.11 ns   | | 1.45 ns   |

As a bonus, here is compared the unit quaternion random implementation from `Eigen` and the one proposed in #105. Note that currently `manif` still uses the implementation from `Eigen` as I don't see a strong motivation to introduce some new code to support for a non critical function. Rather than changing `manif` implementation I'd suggest to propose the change directly upstream to `Eigen`.

|  | Operation  |       | Before |       | After |
| :--- |   :---:   |   :---:   | :---: | :---: | :---: |
|*quaternion*|random|||||
||              | | 524 ns    | | 491 ns    |
||              | | 523 ns    | | 490 ns    |
||              | | 2.04 ns   | | 1.99 ns   |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
+        build_type: [Release, Debug]
         compiler: [{
           "cc": "gcc",
           "cxx": "g++"
@@ -45,7 +46,7 @@ jobs:
           echo "Eigen:" && apt-cache policy libeigen3-dev | grep Installed
       - name: Configure
         working-directory: ${{runner.workspace}}/build
-        run: cmake $GITHUB_WORKSPACE -DBUILD_EXAMPLES=ON -DBUILD_TESTING=ON
+        run: cmake $GITHUB_WORKSPACE -DBUILD_EXAMPLES=ON -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
       - name: Build
         working-directory: ${{runner.workspace}}/build
         run: make -j2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ endif()
 option(BUILD_TESTING "Build all tests." OFF)
 option(ENABLE_CPPCHECK "Enable cppcheck." OFF)
 option(BUILD_EXAMPLES "Build all examples." OFF)
+option(BUILD_BENCHMARKING "Build all benchmarks." OFF)
 option(BUILD_PYTHON_BINDINGS "Build Python bindings with pybind11." OFF)
 option(BUILD_TESTING_PYTHON "Build Python tests only." OFF)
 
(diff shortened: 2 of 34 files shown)


artivis:
Full disclaimer, I did not actually run the math for `SE_2_3Tangent::rjacinv/ljacinv`. I naively/straightforwardly applied the pattern I saw in `SE3Tangent::rjacinv/ljacinv` and the tests passed :)

joansola:
In what consist the optimizations on ljacinv, rjacinv, etc?

artivis:
> In what consist the optimizations on ljacinv, rjacinv, etc?

`SE2Tangent::rjacinv/ljacinv` and `SE_2_3Tangent::rjacinv/ljacinv`?
Well before they were implemented as the inverse of the respective Jacs (e.g. `se2.rjac().inverse()`) whereas this PR actually implements their analytical closed-form (straight to `se2.rjacinv()`).

joansola:
Oh OK I see. Yep, obviously much faster!

artivis:
@joansola This PR is ready when you are. Nevermind the failure coverage test nor the pending ones (looks like a bug).

joansola (approved):
OK I already had a look to all this some time ago, I approve right away

Issue #228, opened 2021-05-05
Distribute Python `sdist` through PyPI


diegoferigo:
I'm very glad to see that Python support https://github.com/artivis/manif/issues/187 has been introduced recently in https://github.com/artivis/manif/pull/201 (previous https://github.com/artivis/manif/pull/179).

I'm opening this issue to discuss / propose the possibility to support distributing the project through PyPI. For those that are not very familiar with the process, there are few incremental steps that could be done, each of them extending more and more the support. To begin with, I'd point interested readers to [https://packaging.python.org/tutorials/packaging-projects/](https://packaging.python.org/tutorials/packaging-projects/) and [Python Packaging User Guide](https://packaging.python.org/).

The main background to understand this issue is [grasping the differences](https://packaging.python.org/guides/distributing-packages-using-setuptools/#packaging-your-project) between a source distribution (aka `sdist`) and a binary distribution (aka wheel or `bdist_wheel`).

As first step, it would be nice distributing `manifpy` in the form of a source distribution. If the project would introduce the support of [`pypa/setuptools_scm`](https://github.com/pypa/setuptools_scm/), the source distribution would be no more no less than a compressed version of the git repository. When users execute `pip install manif`, it would download the `sdist` and run `setup.py`. The process is quite straightforward.

Of course, it is clear to see that in this case, and as documented in https://github.com/artivis/manif/pull/201, `setup.py` expects to find some dependency in the system (eigen, cmake, pybind). Here come [PEP517 and PEP518](https://grassfedcode.medium.com/pep-517-and-518-in-plain-english-47208ca8b7a6) to help us. In few words, during the `pip install manif` process, they create an isolated Python environment (e.g. in `/tmp`) and install the build dependencies before building the `sdist` downloaded from PyPI. In this way, cmake and pybind can be automatically installed, and users do not have to manually install them in their system. AFAIK, eigen is not in PyPI, therefore for the moment it needs to be found in the system.

A very useful tool to implement without much effort all what I explained above is [`diegoferigo/cmake-build-extension`](https://github.com/diegoferigo/cmake-build-extension/), that acts as a bridge between CMake-based projects and setup.py. Disclaimer: I'm the author of the package. Beyond the official documentation, a MWE that also supports PEP600 wheels and implements an automatic CI/CD pipeline is [`diegoferigo/gazebo-yarp-synchronizer`](https://github.com/diegoferigo/gazebo-yarp-synchronizer).

[...]

PR #233, opened 2021-06-16
Update CMake and setuptools logic for the Python bindings

diegoferigo:
This PR refactors both the CMake project and the setuptools files as follows:

- Introduces support to install the Python bindings from the CMake project
- The default destination is inside the `CMAKE_INSTALL_PREFIX`, and the path is computed [as recommended](https://cmake.org/cmake/help/v3.15/module/FindPython3.html#result-variables) for third-party platform dependent projects
- A new optional `DETECT_ACTIVE_PYTHON_SITEPACKAGES` CMake option allows to install the bindings in the active python directory (this means outside the install prefix), that could be a system folder
- The `setup.py` has been simplified and most of its logic moved to a modern `setup.cfg`
- A new `pyproject.toml` was added to support PEP517 and PEP518
- The CMake project is now built with [cmake-build-extension](https://github.com/diegoferigo/cmake-build-extension)
- When packaging with setuptools, if a recent pip version is found, pybind11 is installed in a temporary directory and the user does not have to install it manually beforehand. Then, cmake-build-extension will make sure that manif's CMake finds it.
- Alternatively to pip, also [pypa/build](https://github.com/pypa/build) can be used to generate the sdist / wheel packages
- The pytest configuration was moved to the unified `setup.cfg` file

Related to #228. However, this PR does not implement any CI/CD pipeline towards PyPI.

Test this PR with:

```
docker run -it --rm ubuntu:focal bash

# Install dependencies
apt update
apt install -y build-essential git python3-pip python3-venv libeigen3-dev

# Install pypa/build
python3 -m pip install pip --upgrade
pip3 install build

# Clone the project
cd /tmp
git clone https://github.com/diegoferigo/manif -b feature/setuptools

# Create the sdist and wheel
cd manif
python3 -m build -o dist
```

<details>
<summary>Building the wheel: output</summary>

```
root@f9104509b46a:/tmp/manif# python3 -m build -o dist -w
Found existing installation: setuptools 44.0.0
Uninstalling setuptools-44.0.0:
  Successfully uninstalled setuptools-44.0.0
Collecting setuptools>=45
  Using cached setuptools-57.0.0-py3-none-any.whl (821 kB)
Collecting pybind11
  Using cached pybind11-2.6.2-py2.py3-none-any.whl (191 kB)
Collecting wheel
  Using cached wheel-0.36.2-py2.py3-none-any.whl (35 kB)
Collecting setuptools_scm[toml]>=6.0
  Using cached setuptools_scm-6.0.1-py3-none-any.whl (27 kB)
Collecting cmake-build-extension
  Using cached cmake_build_extension-0.2.0-py3-none-any.whl (8.7 kB)
Collecting cmake>=3.18.2
  Using cached cmake-3.20.3-py2.py3-none-manylinux_2_5_x86_64.manylinux1_x86_64.whl (19.4 MB)
Collecting ninja
  Using cached ninja-1.10.0.post2-py3-none-manylinux1_x86_64.whl (107 kB)
Collecting toml; extra == "toml"
  Using cached toml-0.10.2-py2.py3-none-any.whl (16 kB)
Installing collected packages: setuptools, pybind11, wheel, toml, setuptools-scm, cmake-build-extension, cmake, ninja
Successfully installed cmake-3.20.3 cmake-build-extension-0.2.0 ninja-1.10.0.post2 pybind11-2.6.2 setuptools-57.0.0 setuptools-scm-6.0.1 toml-0.10.2 wheel-0.36.2
running egg_info
writing manifpy.egg-info/PKG-INFO
writing dependency_links to manifpy.egg-info/dependency_links.txt
writing requirements to manifpy.egg-info/requires.txt
writing top-level names to manifpy.egg-info/top_level.txt
adding license file 'LICENSE'
writing manifest file 'manifpy.egg-info/SOURCES.txt'
Requirement already satisfied: wheel in /tmp/build-env-gtbc_2p5/lib/python3.8/site-packages (from -r /tmp/build-reqs-lxpoow59.txt (line 1)) (0.36.2)
running bdist_wheel
running build
running build_ext

==> Configuring:
$ cmake -S /tmp/manif -B /tmp/manif/build/temp.linux-x86_64-3.8_CMakeProject -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX:PATH=/tmp/manif/build/lib.linux-x86_64-3.8/manifpy -DCALL_FROM_SETUP_PY:BOOL=ON -DBUILD_PYTHON_BINDINGS:BOOL=ON

==> Building:
$ cmake --build /tmp/manif/build/temp.linux-x86_64-3.8_CMakeProject --config Release

==> Installing:
$ cmake --install /tmp/manif/build/temp.linux-x86_64-3.8_CMakeProject

-- The CXX compiler identification is GNU 9.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test COMPILER_SUPPORTS_CXX11
-- Performing Test COMPILER_SUPPORTS_CXX11 - Success
-- The compiler /usr/bin/c++ has C++11 support.
CMake Warning at CMakeLists.txt:50 (message):
  Eigen3 version 3.3.7 found in /usr/include/eigen3.  Beware that the move
  semantic has a bug and resolves to a simple copy.


-- Found Python: /usr/bin/python3.8 (found version "3.8.5") found components: Interpreter Development Development.Module Development.Embed 
-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Success
-- Found pybind11: /tmp/build-env-gtbc_2p5/lib/python3.8/site-packages/pybind11/include (found version "2.6.2" )
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/manif/build/temp.linux-x86_64-3.8_CMakeProject
[8/8] Linking CXX shared module manifpy/bindings/manifpy.cpython-38-x86_64-linux-gnu.so
-- Install configuration: "Release"
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/share/manif/cmake/manifTargets.cmake
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/share/manif/cmake/manifConfigVersion.cmake
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/share/manif/cmake/manifConfig.cmake
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/tl/optional.hpp
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/SE_2_3.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/functions.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2_properties.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2Tangent_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2Tangent.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2Tangent_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/random.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/traits.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2Tangent_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2Tangent.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2Tangent_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2_properties.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/lie_group_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/assignment_assert.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3Tangent_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3Tangent.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3Tangent_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3_properties.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/utils.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/macro.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3_properties.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3Tangent_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3Tangent.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3Tangent_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/Rn_properties.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/Rn.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/Rn_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/RnTangent.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/RnTangent_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/Rn_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/RnTangent_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/generator.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3Tangent_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3Tangent_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3Tangent.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3_map.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3_properties.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/eigen.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/tangent_base.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/Rn.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/SO2.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/SO3.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/SE2.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/ceres.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/local_parametrization.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/ceres_utils.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/objective.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/constants.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/constraint.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/SE3.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/constants.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/manif.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms/average.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms/interpolation.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms/bezier.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms/decasteljau.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/gtest/gtest_eigen_utils.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/include/manif/gtest/gtest_manif_utils.h
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/bindings/manifpy.cpython-38-x86_64-linux-gnu.so
-- Installing: /tmp/manif/build/lib.linux-x86_64-3.8/manifpy/__init__.py
installing to build/bdist.linux-x86_64/wheel
running install
running install_lib
creating build/bdist.linux-x86_64
creating build/bdist.linux-x86_64/wheel
creating build/bdist.linux-x86_64/wheel/manifpy
creating build/bdist.linux-x86_64/wheel/manifpy/bindings
copying build/lib.linux-x86_64-3.8/manifpy/bindings/manifpy.cpython-38-x86_64-linux-gnu.so -> build/bdist.linux-x86_64/wheel/manifpy/bindings
creating build/bdist.linux-x86_64/wheel/manifpy/include
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/SE_2_3.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/functions.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2_properties.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2Tangent_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2Tangent.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2Tangent_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se2/SE2_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/random.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/traits.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2Tangent_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2Tangent.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2Tangent_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so2/SO2_properties.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so2
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/lie_group_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/assignment_assert.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3Tangent_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3Tangent.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3Tangent_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/so3/SO3_properties.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/so3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/utils.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/macro.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se_2_3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se_2_3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se_2_3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3_properties.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se_2_3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se_2_3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3Tangent_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se_2_3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3Tangent.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se_2_3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se_2_3/SE_2_3Tangent_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se_2_3
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/rn
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/Rn_properties.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/rn
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/Rn.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/rn
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/Rn_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/rn
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/RnTangent.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/rn
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/RnTangent_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/rn
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/Rn_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/rn
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/rn/RnTangent_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/rn
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/generator.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3Tangent_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3Tangent_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3Tangent.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3_map.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/se3/SE3_properties.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl/se3
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/eigen.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/impl/tangent_base.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/impl
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/Rn.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/SO2.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/SO3.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/SE2.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/ceres
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/ceres.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/ceres
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/local_parametrization.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/ceres
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/ceres_utils.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/ceres
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/objective.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/ceres
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/constants.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/ceres
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/ceres/constraint.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/ceres
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/SE3.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/constants.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/manif.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/gtest
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/gtest/gtest_manif_utils.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/gtest
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/gtest/gtest_eigen_utils.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/gtest
creating build/bdist.linux-x86_64/wheel/manifpy/include/manif/algorithms
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms/average.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/algorithms
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms/interpolation.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/algorithms
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms/bezier.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/algorithms
copying build/lib.linux-x86_64-3.8/manifpy/include/manif/algorithms/decasteljau.h -> build/bdist.linux-x86_64/wheel/manifpy/include/manif/algorithms
creating build/bdist.linux-x86_64/wheel/manifpy/include/tl
copying build/lib.linux-x86_64-3.8/manifpy/include/tl/optional.hpp -> build/bdist.linux-x86_64/wheel/manifpy/include/tl
creating build/bdist.linux-x86_64/wheel/manifpy/share
creating build/bdist.linux-x86_64/wheel/manifpy/share/manif
creating build/bdist.linux-x86_64/wheel/manifpy/share/manif/cmake
copying build/lib.linux-x86_64-3.8/manifpy/share/manif/cmake/manifConfigVersion.cmake -> build/bdist.linux-x86_64/wheel/manifpy/share/manif/cmake
copying build/lib.linux-x86_64-3.8/manifpy/share/manif/cmake/manifTargets.cmake -> build/bdist.linux-x86_64/wheel/manifpy/share/manif/cmake
copying build/lib.linux-x86_64-3.8/manifpy/share/manif/cmake/manifConfig.cmake -> build/bdist.linux-x86_64/wheel/manifpy/share/manif/cmake
copying build/lib.linux-x86_64-3.8/manifpy/__init__.py -> build/bdist.linux-x86_64/wheel/manifpy
running install_egg_info
running egg_info
writing manifpy.egg-info/PKG-INFO
writing dependency_links to manifpy.egg-info/dependency_links.txt
writing requirements to manifpy.egg-info/requires.txt
writing top-level names to manifpy.egg-info/top_level.txt
adding license file 'LICENSE'
writing manifest file 'manifpy.egg-info/SOURCES.txt'
Copying manifpy.egg-info to build/bdist.linux-x86_64/wheel/manifpy-0.0.5.dev3-py3.8.egg-info
running install_scripts
adding license file "LICENSE" (matched pattern "LICEN[CS]E*")
creating build/bdist.linux-x86_64/wheel/manifpy-0.0.5.dev3.dist-info/WHEEL
creating '/tmp/manif/dist/tmp8oi49s20/manifpy-0.0.5.dev3-cp38-cp38-linux_x86_64.whl' and adding 'build/bdist.linux-x86_64/wheel' to it
adding 'manifpy/__init__.py'
adding 'manifpy/bindings/manifpy.cpython-38-x86_64-linux-gnu.so'
adding 'manifpy/include/manif/Rn.h'
adding 'manifpy/include/manif/SE2.h'
adding 'manifpy/include/manif/SE3.h'
adding 'manifpy/include/manif/SE_2_3.h'
adding 'manifpy/include/manif/SO2.h'
adding 'manifpy/include/manif/SO3.h'
adding 'manifpy/include/manif/constants.h'
adding 'manifpy/include/manif/functions.h'
adding 'manifpy/include/manif/manif.h'
adding 'manifpy/include/manif/algorithms/average.h'
adding 'manifpy/include/manif/algorithms/bezier.h'
adding 'manifpy/include/manif/algorithms/decasteljau.h'
adding 'manifpy/include/manif/algorithms/interpolation.h'
adding 'manifpy/include/manif/ceres/ceres.h'
adding 'manifpy/include/manif/ceres/ceres_utils.h'
adding 'manifpy/include/manif/ceres/constants.h'
adding 'manifpy/include/manif/ceres/constraint.h'
adding 'manifpy/include/manif/ceres/local_parametrization.h'
adding 'manifpy/include/manif/ceres/objective.h'
adding 'manifpy/include/manif/gtest/gtest_eigen_utils.h'
adding 'manifpy/include/manif/gtest/gtest_manif_utils.h'
adding 'manifpy/include/manif/impl/assignment_assert.h'
adding 'manifpy/include/manif/impl/eigen.h'
adding 'manifpy/include/manif/impl/generator.h'
adding 'manifpy/include/manif/impl/lie_group_base.h'
adding 'manifpy/include/manif/impl/macro.h'
adding 'manifpy/include/manif/impl/random.h'
adding 'manifpy/include/manif/impl/tangent_base.h'
adding 'manifpy/include/manif/impl/traits.h'
adding 'manifpy/include/manif/impl/utils.h'
adding 'manifpy/include/manif/impl/rn/Rn.h'
adding 'manifpy/include/manif/impl/rn/RnTangent.h'
adding 'manifpy/include/manif/impl/rn/RnTangent_base.h'
adding 'manifpy/include/manif/impl/rn/RnTangent_map.h'
adding 'manifpy/include/manif/impl/rn/Rn_base.h'
adding 'manifpy/include/manif/impl/rn/Rn_map.h'
adding 'manifpy/include/manif/impl/rn/Rn_properties.h'
adding 'manifpy/include/manif/impl/se2/SE2.h'
adding 'manifpy/include/manif/impl/se2/SE2Tangent.h'
adding 'manifpy/include/manif/impl/se2/SE2Tangent_base.h'
adding 'manifpy/include/manif/impl/se2/SE2Tangent_map.h'
adding 'manifpy/include/manif/impl/se2/SE2_base.h'
adding 'manifpy/include/manif/impl/se2/SE2_map.h'
adding 'manifpy/include/manif/impl/se2/SE2_properties.h'
adding 'manifpy/include/manif/impl/se3/SE3.h'
adding 'manifpy/include/manif/impl/se3/SE3Tangent.h'
adding 'manifpy/include/manif/impl/se3/SE3Tangent_base.h'
adding 'manifpy/include/manif/impl/se3/SE3Tangent_map.h'
adding 'manifpy/include/manif/impl/se3/SE3_base.h'
adding 'manifpy/include/manif/impl/se3/SE3_map.h'
adding 'manifpy/include/manif/impl/se3/SE3_properties.h'
adding 'manifpy/include/manif/impl/se_2_3/SE_2_3.h'
adding 'manifpy/include/manif/impl/se_2_3/SE_2_3Tangent.h'
adding 'manifpy/include/manif/impl/se_2_3/SE_2_3Tangent_base.h'
adding 'manifpy/include/manif/impl/se_2_3/SE_2_3Tangent_map.h'
adding 'manifpy/include/manif/impl/se_2_3/SE_2_3_base.h'
adding 'manifpy/include/manif/impl/se_2_3/SE_2_3_map.h'
adding 'manifpy/include/manif/impl/se_2_3/SE_2_3_properties.h'
adding 'manifpy/include/manif/impl/so2/SO2.h'
adding 'manifpy/include/manif/impl/so2/SO2Tangent.h'
adding 'manifpy/include/manif/impl/so2/SO2Tangent_base.h'
adding 'manifpy/include/manif/impl/so2/SO2Tangent_map.h'
adding 'manifpy/include/manif/impl/so2/SO2_base.h'
adding 'manifpy/include/manif/impl/so2/SO2_map.h'
adding 'manifpy/include/manif/impl/so2/SO2_properties.h'
adding 'manifpy/include/manif/impl/so3/SO3.h'
adding 'manifpy/include/manif/impl/so3/SO3Tangent.h'
adding 'manifpy/include/manif/impl/so3/SO3Tangent_base.h'
adding 'manifpy/include/manif/impl/so3/SO3Tangent_map.h'
adding 'manifpy/include/manif/impl/so3/SO3_base.h'
adding 'manifpy/include/manif/impl/so3/SO3_map.h'
adding 'manifpy/include/manif/impl/so3/SO3_properties.h'
adding 'manifpy/include/tl/optional.hpp'
adding 'manifpy/share/manif/cmake/manifConfig.cmake'
adding 'manifpy/share/manif/cmake/manifConfigVersion.cmake'
adding 'manifpy/share/manif/cmake/manifTargets.cmake'
adding 'manifpy-0.0.5.dev3.dist-info/LICENSE'
adding 'manifpy-0.0.5.dev3.dist-info/METADATA'
adding 'manifpy-0.0.5.dev3.dist-info/WHEEL'
adding 'manifpy-0.0.5.dev3.dist-info/top_level.txt'
adding 'manifpy-0.0.5.dev3.dist-info/RECORD'
removing build/bdist.linux-x86_64/wheel
```

</details>


artivis:
Hi,
Thanks for the contribution, looks a lot simpler than what I was tinkering with :+1:

diegoferigo:
> Hi,
> Thanks for the contribution, looks a lot simpler than what I was tinkering with +1

Great! I'm going to debug the failure in CI, I'm not sure why `pip install .` fails there. Locally it does not give me any trouble.

diegoferigo:
Mmh Github is blocking CI since this is my first contribution. @artivis I suspect you have to manually approve every run. I've just enabled Actions in my fork, I'll do a force push to trigger the build and continue in this way until I solve the problem.

artivis (commented):
> Mmh Github is blocking CI since this is my first contribution. @artivis I suspect you have to manually approve every run. I've just enabled Actions in my fork, I'll do a force push to trigger the build and continue in this way until I solve the problem.

I had approved the run already this morning. Could it be because of the force-push?

pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[build-system]
+requires = [
+    "wheel",
+    "setuptools>=45",
+    "setuptools_scm[toml]>=6.0",
+    "ninja",
+    "cmake>=3.18.2",
```
artivis, line 7: Mind that `manif` support goes back to Ubuntu 18.04 which [ships with CMake 3.10.2](https://packages.ubuntu.com/search?keywords=cmake).

CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ option(BUILD_PYTHON_BINDINGS "Build Python bindings with pybind11." OFF)
 option(BUILD_TESTING_PYTHON "Build Python tests only." OFF)
 
 if (BUILD_PYTHON_BINDINGS)
+  set(PYBIND11_FINDPYTHON ON)
```
artivis, line 64: According to [the doc](https://pybind11.readthedocs.io/en/stable/cmake/index.html#modes), this is a CMake 3.12 (3.15 recommended) feature. `manif` support goes back to Ubuntu 18.04 which [ships with CMake 3.10.2](https://packages.ubuntu.com/search?keywords=cmake).

pyproject.toml
diegoferigo: Yes, this is correct. However, this is the CMake version that is installed in the virtualenv created automatically during the wheel build process. I used this version on purpose since it's [the recommendation from pybind11](https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode). Such CMake versions include a much better Python logic, and since this does not affect in any case the user's setup, I used a more recent require version. In most cases, at the time on writing, the build process will pull CMake >= 3.20.

CMakeLists.txt
diegoferigo: Right, I missed this. I guess in this case I'm going to revert to the classic mode. The only problem is that I need to get the location of python interpreter. I'll figure out what is the best way with CMake 3.12.

An alternative I used in other projects is vendoring a modern FindPython files in the project, so that the behavior is consistent among all OSs and CMake versions. If you prefer proceeding this way, I can update this PR accordingly.

diegoferigo: Btw, I just noticed that in Ubuntu 16.04, python2 is found. Not sure if this is the intended behavior:

https://github.com/diegoferigo/manif/runs/2839281906?check_suite_focus=true#step:5:85

artivis: > An alternative I used in other projects is vendoring a modern FindPython files in the project, so that the behavior is consistent among all OSs and CMake versions. If you prefer proceeding this way, I can update this PR accordingly.

I'd actually prefer to vendor as little as possible since that means more things to maintain. Especially true with things I'm not familiar with.

diegoferigo: > > An alternative I used in other projects is vendoring a modern FindPython files in the project, so that the behavior is consistent among all OSs and CMake versions. If you prefer proceeding this way, I can update this PR accordingly.
> 
> I'd actually prefer to vendor as little as possible since that means more things to maintain. Especially true with things I'm not familiar with.

I understand it, no problem. Right now, in order to debug the failing test, I vendored the resources in dff4930. This removes from the picture possible problems of CMake detection. It seems that the error in CI occurrs only with Python 3.6, if I update the pybind11 job to 3.8 it succeeds. Not yet sure why, maybe a linked library is missing.

Anyway, the trade-off between vendoring or using plain CMake, is that old CMake version have either poor or no Python support at all. CMake 3.10, for instance, only has [`FindPythonInterp`](https://cmake.org/cmake/help/v3.10/module/FindPythonInterp.html) that in my experience can create many mismatch in the found Python version. Though, if you prefer dealing with possible issues rather than maintaining the vendored files, I will update the logic to support plain CMake 3.10.

artivis: > Though, if you prefer dealing with possible issues rather than maintaining the vendored files

Well when you put it that way I'm not sure I really have a choice :sweat_smile:. 
Again, my concern is that I'm not familiar with all that and I care about being able to maintain the feature without too much pain.

diegoferigo:
I'm not being able to reproduce the failures in importing the bindings that occur for Python <= 3.7. Locally, on a ubuntu 16.04 with Python from ppa, everything seems running fine. Only CI is giving troubles. Ideas?

CMakeLists.txt
diegoferigo: In the past I had to deal with a lot of related issues and I'm biased about it :) The vendored files come directly from the [CMake tree](https://github.com/Kitware/CMake/tree/v3.20.4/Modules), and unless major changes (which I doubt will happen) these vendored resources can stay as they are for years. I'd not be very concerned about their maintenance.

I managed to solve the problems in CI by passing `Python3_EXECUTABLE` as was done before. Even though in other projects I didn't have the need to use this hint, in this project for some reason it's the only way to have a proper detection from `setup.py` for all the supported platforms (ubuntu and macos) and all the python versions (3.6 - 3.9).

diegoferigo:
The PR seems ok now. Few final remarks not included in the first post:

- This PR changes the structure of the python bindings from a single file (the shared library generated by pybind11) to a proper package that consist of a `manifpy` folder with inside the pybind11 shared library and a `__init__.py` file. The downstream usage is not affected thanks to the following pattern used in the `__init__.py`: `from _bindings import *`.
- I enabled multiple Python versions in the `pybind11` CI job.
- I also enabled the upload of the Python artifacts (sdist and wheels) for all the supported platforms and python versions.

.github/workflows/ci.yml
```diff
@@ -214,12 +214,13 @@ jobs:
...
         platform: [macos-latest, ubuntu-latest] #windows-latest,
-        # python-version: [3.5, 3.6, 3.7, 3.8]
-        python-version: [3.6]
+        python-version: [3.6, 3.7, 3.8, 3.9]
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
artivis, line 223: Is that necessary?

python/bindings_manif.cpp
```diff
@@ -10,7 +10,7 @@ void wrap_SE3(pybind11::module &m);
 
 void wrap_SE_2_3(pybind11::module &m);
 
-PYBIND11_MODULE(manifpy, m) {
+PYBIND11_MODULE(_bindings, m) {
```
artivis, line 13: Genuine question, why the renaming? 
Altho unlikely, `_binding` may be a little too generic and cause conflicts. How about `_manifpy`, `_manif_bindings` or such?

artivis (changes requested):
Thanks a bunch for the effort!
There are still a few things I need to figure out, I'll give it another later on.

In the meantime, can you please move the new vendored cmake files to the `cmake/module/` sub-folder?

CMakeLists.txt
```diff
@@ -60,7 +60,18 @@ option(BUILD_BENCHMARKING "Build all benchmarks." OFF)
...
+list(PREPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+# Find a suitable Python installation
+find_package(Python3 COMPONENTS Interpreter Development.Module)
+
 if (BUILD_PYTHON_BINDINGS)
+
+  if (NOT Python3_FOUND)
+    message(FATAL_ERROR "Failed to find a compatible Python3 installation")
+  endif()
```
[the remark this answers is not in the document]
diegoferigo: > Could this be rearranged so that,
> 
>     * we try to find Python3 only if necessary

I proceeded with the current implementation because Python is needed 1) when the bindings are created and 2) when the Python tests are enabled. The find logic can be guarded with the flags used to include these two subfolders.
 
>     * we first try to find Python3 with the system's CMake modules and only use the vendored one if it fails - similarly to Eigen [above](https://github.com/artivis/manif/blob/cd1d1403430ce56ae1d371eb82f3e287c9a4e82f/CMakeLists.txt#L28)??

I guess something similar could be done, even if I don't see the point of not using the vendored files all the times since they're there. The only problem I see in using system's modules first is that often a Python interpreter is detected among the many options in the system, but is the wrong one, or one that does not have all the components. Therefore, I see the fallback method as not very robust. The worse case is when the interpreter and the libs are detected with different versions, and this could happen in very old CMake versions. Just to understand your point, what kind of problem do you expect when only the vendored files are used? It seems to me a good practice to provide a more uniform behavior without complicating the logic.

.github/workflows/ci.yml
diegoferigo: This PR uses [pypa/setuptools_scm](https://github.com/pypa/setuptools_scm) to 1) create the sdist and 2) automatically detect the package version from the latest tag (compliant with PEP440).

This means that the git clone must have all the history including the tags. However, inspecting the artifacts I see that `0.1.0.devX` is used. I typically use [another pattern](https://github.com/robotology/idyntree/blob/ab9336447ab0c6d6d9f9d0c3d3a19be31140aed3/.github/workflows/python.yml#L73-L74), I'll try to apply it also here and check if the versioning get fixed.

python/bindings_manif.cpp
diegoferigo: This is related to the first point of https://github.com/artivis/manif/pull/233#issuecomment-862604986.

Referring to the installed tree in the active site-packages, now the bindings file is no longer the main entry point of the package. In fact, now a new `manifpy` folder is created and the bindings file (the cpython shared lib) is installed inside it.

Before this PR, manifpy could be used as:

```python
import manifpy
from manifpy import SE3

se2 = manifpy.SE2()
se3 = SE3()
```

This PR is backward compatible with that usage because the `__init__.py` file contains `from _bindings import *`. This also means that dowstream users, in addition to the snippet avobe, can also do the following:

```
import manifpy._bindings

se2 = manifpy._bindings.SE2()
se3 = manifpy._bindings.SE3()
```

The leading `_` in python usually means a private item, and while the last snippet is totally legit, it implies that the real C module is an implementation detail. It can have any name you like, I don't have strong opinions on it.

In light of the explanation above, please let me know the name you prefer. I could even revert to the original name that would result in the following valid snippet:

```
import manifpy
from manifpy import SE3
import manifpy.manifpy

se2 = manifpy.SE2()
se3 = SE3()
so3 = manifpy.manifpy.SO3()
```

I like using `bindings` or `_bindings` for the real cpython module, but it's just personal taste and there is no common pattern I noticed around.

<details>
<summary>Package Tree</summary>

```
manif on  feature/setuptools via △ v3.20.2 🅒 /conda  
❯ tree /conda/lib/python3.8/site-packages/manifpy
/conda/lib/python3.8/site-packages/manifpy
├── __init__.py
├── _bindings.cpython-38-x86_64-linux-gnu.so
├── include
│   ├── manif
│   │   ├── Rn.h
│   │   ├── SE2.h
│   │   ├── SE3.h
│   │   ├── SE_2_3.h
│   │   ├── SO2.h
│   │   ├── SO3.h
│   │   ├── algorithms
│   │   │   ├── average.h
│   │   │   ├── bezier.h
│   │   │   ├── decasteljau.h
│   │   │   └── interpolation.h
│   │   ├── ceres
│   │   │   ├── ceres.h
│   │   │   ├── ceres_utils.h
│   │   │   ├── constants.h
│   │   │   ├── constraint.h
│   │   │   ├── local_parametrization.h
│   │   │   └── objective.h
│   │   ├── constants.h
│   │   ├── functions.h
│   │   ├── gtest
│   │   │   ├── gtest_eigen_utils.h
│   │   │   └── gtest_manif_utils.h
│   │   ├── impl
│   │   │   ├── assignment_assert.h
│   │   │   ├── eigen.h
│   │   │   ├── generator.h
│   │   │   ├── lie_group_base.h
│   │   │   ├── macro.h
│   │   │   ├── random.h
│   │   │   ├── rn
│   │   │   │   ├── Rn.h
│   │   │   │   ├── RnTangent.h
│   │   │   │   ├── RnTangent_base.h
│   │   │   │   ├── RnTangent_map.h
│   │   │   │   ├── Rn_base.h
│   │   │   │   ├── Rn_map.h
│   │   │   │   └── Rn_properties.h
│   │   │   ├── se2
│   │   │   │   ├── SE2.h
│   │   │   │   ├── SE2Tangent.h
│   │   │   │   ├── SE2Tangent_base.h
│   │   │   │   ├── SE2Tangent_map.h
│   │   │   │   ├── SE2_base.h
│   │   │   │   ├── SE2_map.h
│   │   │   │   └── SE2_properties.h
│   │   │   ├── se3
│   │   │   │   ├── SE3.h
│   │   │   │   ├── SE3Tangent.h
│   │   │   │   ├── SE3Tangent_base.h
│   │   │   │   ├── SE3Tangent_map.h
│   │   │   │   ├── SE3_base.h
│   │   │   │   ├── SE3_map.h
│   │   │   │   └── SE3_properties.h
│   │   │   ├── se_2_3
│   │   │   │   ├── SE_2_3.h
│   │   │   │   ├── SE_2_3Tangent.h
│   │   │   │   ├── SE_2_3Tangent_base.h
│   │   │   │   ├── SE_2_3Tangent_map.h
│   │   │   │   ├── SE_2_3_base.h
│   │   │   │   ├── SE_2_3_map.h
│   │   │   │   └── SE_2_3_properties.h
│   │   │   ├── so2
│   │   │   │   ├── SO2.h
│   │   │   │   ├── SO2Tangent.h
│   │   │   │   ├── SO2Tangent_base.h
│   │   │   │   ├── SO2Tangent_map.h
│   │   │   │   ├── SO2_base.h
│   │   │   │   ├── SO2_map.h
│   │   │   │   └── SO2_properties.h
│   │   │   ├── so3
│   │   │   │   ├── SO3.h
│   │   │   │   ├── SO3Tangent.h
│   │   │   │   ├── SO3Tangent_base.h
│   │   │   │   ├── SO3Tangent_map.h
│   │   │   │   ├── SO3_base.h
│   │   │   │   ├── SO3_map.h
│   │   │   │   └── SO3_properties.h
│   │   │   ├── tangent_base.h
│   │   │   ├── traits.h
│   │   │   └── utils.h
│   │   └── manif.h
│   └── tl
│       └── optional.hpp
└── share
    └── manif
        └── cmake
            ├── manifConfig.cmake
            ├── manifConfigVersion.cmake
            └── manifTargets.cmake
```

</details>

CMakeLists.txt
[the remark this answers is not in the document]
diegoferigo: Could 2620291 be ok?

CMakeLists.txt
```diff
@@ -300,6 +311,10 @@ endif(BUILD_TESTING)
 
 if(BUILD_PYTHON_BINDINGS AND (BUILD_TESTING OR BUILD_TESTING_PYTHON))
 
+  if (NOT Python3_FOUND)
+    message(FATAL_ERROR "Failed to find a compatible Python3 installation")
+  endif()
+
```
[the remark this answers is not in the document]
diegoferigo: Done in 2620291

.github/workflows/ci.yml
diegoferigo: Update in c2a9678, I will check the new artifacts if the version is fixed.

diegoferigo:
> In the meantime, can you please move the new vendored cmake files to the `cmake/module/` sub-folder?

Done in 533a1cf. I'll squash it with the other commit before merging.

> Edit: Can you also update (where necessary) the [Python quick start](https://github.com/diegoferigo/manif/blob/cd1d1403430ce56ae1d371eb82f3e287c9a4e82f/docs/pages/python/Quick-start.md) and the [CONTRIBUTING.md](https://github.com/diegoferigo/manif/blob/cd1d1403430ce56ae1d371eb82f3e287c9a4e82f/CONTRIBUTING.md) pages to reflect those changes?

Done in 3cf0178,  63e8117 and 2413377. The`requirements.txt` file is no longer necessary since tools compatible with PEP517 and PEP518 like pip and pypa/build will read all the information they need from `setup.cfg` and `pyproject.toml`.

The only caveat is for installing the dependencies of the pure CMake project, that is what described in CONTRIBUTING. In this case, the installation of pybind11[global] (that is quite a hack) is still necessary.

Then, can you confirm that `numpy` is only a testing dependency and is not needed during regular runtime?

CMakeLists.txt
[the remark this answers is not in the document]
artivis: Concerning the first point, looks good :+1:.
As for the second, I usually prefer to rely on what's shipped on the system by default, that's how thing should be and we shouldn't need to vendor. However in Eigen case it is simpler, it's binary, `find_package` works or not. In Python3 case, if there is also the possibility of a 'messed up' detection then I guess it make sense to rely solely on the vendored cmake files.

CMakeLists.txt
```diff
@@ -61,8 +61,15 @@ option(BUILD_PYTHON_BINDINGS "Build Python bindings with pybind11." OFF)
 option(BUILD_TESTING_PYTHON "Build Python tests only." OFF)
 
 if (BUILD_PYTHON_BINDINGS)
+
+  # Include the vendored FindPython3 resources
+  list(PREPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
+
+  find_package(Python3 COMPONENTS Interpreter Development.Module)
```
artivis, line 68: I think you should use `REQUIRED` in `find_package`: 
"The REQUIRED option stops processing with an error message if the package cannot be found."

```suggestion
  find_package(Python3 REQUIRED COMPONENTS Interpreter Development.Module)
```

python/bindings_manif.cpp
artivis: Thanks for the detailed explanation. Let's leave it as it is, I guess there is no actual chance for a naming conflict.

CMakeLists.txt
diegoferigo: Good catch, thanks! I copied and pasted this line from the previous logic that didn't require REQUIRED and then forgot to update it.

diegoferigo: I commited and squashed the suggestion, I forgot to get the commit before force pushing but it is now included in 2aea66e.

diegoferigo:
I addressed all the comments and cleaned the history.

There's a last remark I have to make and it regards the branching strategy that is used. The version of the Python package is determined by [pypa/setuptools_scm](https://github.com/pypa/setuptools_scm) from the closest tag found from the HEAD. Currently, `devel` has no parent commit associated with a tag since all the tagging happens in `master.

What I recommend in order to get correct development tags in `devel` is merging `master` into `devel` after a new tag. Referring to https://github.com/robotology/gym-ignition/issues/278, here's a descriptive image:

Screenshot_20201123_131218

CMakeLists.txt
diegoferigo: Done in adf6fe6

artivis:
> There's a last remark I have to make and it regards the branching strategy that is used.

Say no more, I'm well aware it is kind of crappy :sweat_smile: as it makes my own work more complicated. I'll somehow merge both branches asap and possibly get rid of `master`. This should make things easier.

Thanks again for iterating on this PR. 
Unfortunately I'm about to go on a summer break and I'm afraid this will have to wait for ~ a couple weeks. I'd prefer no to rush the merge now and forget about it...

diegoferigo:
> > There's a last remark I have to make and it regards the branching strategy that is used.
> 
> Say no more, I'm well aware it is kind of crappy :sweat_smile:  as it makes my own work more complicated. I'll somehow merge both branches asap and possibly get rid of `master`. This should make things easier.

A single branch with a quick release cycle is a great choice, especially in research :) The logic of the automatic versioning is a bit obscure and indeed makes these branching needs a bit odd, but after a couple of years of usage I have to say that it is handy, especially if you maintain a nightly channel corresponding to new commits landing the development branch (that could be `main`). The release version will be something like `v0.5.0.devX` and it's automatically generated from the git repo.

> Thanks again for iterating on this PR.
> Unfortunately I'm about to go on a summer break and I'm afraid this will have to wait for ~ a couple weeks. I'd prefer no to rush the merge now and forget about it...

Thank you for the feedback and no problem for the delay, have a nice vacation and good rest :wink:

python/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 set(PYBIND11_CPP_STANDARD -std=c++11)
-pybind11_add_module(manifpy
+pybind11_add_module(manifpy MODULE
```
artivis, line 2: Note to myself, `MODULE` is the default when nothing's specified.

artivis:
Hi @diegoferigo,

So I'm back on this PR and gave it a shot yesterday in a clean 18.04 container. Unfortunately the plain CMake build of the Python bindings does not work anymore. It first fails finding Python3 (not sure why since we provide the module) and then on the install rule. I've revisited the FindPython logic and will push that on a separate branch later today to run it through the CI. As for the install rule, I'll open a separate comment.

python/CMakeLists.txt
```diff
@@ -17,3 +17,47 @@ target_compile_definitions(manifpy PRIVATE EIGEN_DEFAULT_TO_ROW_MAJOR)
...
+# Installation in the CMAKE_INSTALL_PREFIX
+else()
+  execute_process(
+    COMMAND
+    ${Python3_EXECUTABLE} -c
+    "import distutils.sysconfig; print(distutils.sysconfig.get_python_lib(plat_specific=True, standard_lib=False, prefix=''))"
+    OUTPUT_VARIABLE _PYTHON_INSTDIR)
+  string(STRIP ${_PYTHON_INSTDIR} _PYTHON_INSTDIR_CLEAN)
+  set(PYTHON_INSTDIR "${_PYTHON_INSTDIR_CLEAN}/manifpy")
+endif()
```
artivis, line 42: So, I don't quite understand this chunk. Correct me if I'm wrong but the overall idea is to choose whether to install in the CMake install path or the Python's site-package right? 

Now the logic seem odd; aren't the two last blocks inverted? Shouldn't it be 
```cmake
elseif(CALL_FROM_SETUP_PY)
  execute_process(...)
else()
  set(PYTHON_INSTDIR "${CMAKE_INSTALL_PREFIX}")
```
??

Furthermore, the variable `Python3_SITELIB` is provided by the FindPython3 module and is the results of `distutils.sysconfig.get_python_lib(plat_specific=True, standard_lib=False)` (see [doc](https://cmake.org/cmake/help/v3.14/module/FindPython3.html)). So why do we have two distinct cases leading to the same result?

I don't know what's the common practice (if any) but shouldn't we install by default in the Python site-package?

artivis: Scratch my previous comment about the 'if' block logic being odd, I went down the rabbit hole :sob:.
Still, the `CMAKE_INSTALL_PREFIX` variable populated by `cmake-build-extension` is anything but a prefix :sweat_smile:.

diegoferigo: Yeah sorry for replying too late, it would have saved you quite some digging :sweat_smile: For the records, in our projects we always find odd when a CMake projects installs files outside `CMAKE_INSTALL_PREFIX`, this is the rationale behind this logic. This is a behavior that could surprise users and third-party packagers and we try to avoid it.

However, in this specific case, using the detected `site-packages` folder (that could be a system folder) makes sense in some use cases. In fact, installing in a custom `CMAKE_INSTALL_PREFIX`, in most cases, requires the user to extend manually `PYTHONPATH` if they are using the system interpreter. Nowadays I believe this is not the most common case and this is the reason why the default behavior of this CMake logic in to install the bindings inside the `CMAKE_INSTALL_PREFIX`, where the right installation path has to be structured. Here is where `distutils` enters in the picture.

artivis: No worries, it's actually a good thing that I investigated as it allowed me to better understand what's going on.
I generally agree with your point on installing things under the `CMAKE_INSTALL_PREFIX` but for the particular case of python bindings installed from plain CMake, I'm not sure if that makes sense and at the very least it is not convenient since it may requires (as you noted) to augment the `PYTHONPATH`. Now, I don't actually expect anyone to use the plain CMake install for the bindings tho, and I'd encourage using `pip`.

.gitignore
```diff
@@ -15,3 +15,4 @@ docs/m.css
 *.so
 *__pycache__
 .pytest_cache
+dist
```
diegoferigo, line 18: On the long term, you can consider using the official gitignore for Python:

https://github.com/github/gitignore/blob/master/Python.gitignore

artivis:
All green! Thanks a lot @diegoferigo for the feature and iterating this PR! Merging :)

PR #234, opened 2021-07-10
Add Bundle and BundleTangent types

### LINKED ISSUE #84: Feature: support for composite groups
In several developments it is handy to consider composite manifolds, such as T3xSO3 instead of SE3, or the IMU implementation using T3xT3xSO3.

In the paper, a notation is introduced so as to be able to define, for composite manifolds, the full family of operators: composition, inverse, exp, log, plus and minus, mainly. It is also possible to produce the Jacobians as a block-concatenation of the jacobians of each individual group.

It'd be very interesting if we could create a (probably variadic templated) way of producing composites of any number of groups, and then have the API ready to manipulate them as a single group.

This is probably not trivial to do.

artivis:
Replaces #175.
See #175 comments for a discussion around the feature.

Closes #84.
Closes #175.

## Notes

- Ubuntu 16.04 is deprecated in CI. The distro is EOL and it's support will be soon dropped in github actions (that's fortunate because this PR causes a compiler internal error on 16.04)
- Bundle is **disabled** on MSVC-based CI. Looks like MSVC is full of issues with alias-type and it doesn't like the Rn alias (R1/R2/R3 etc...)

# Original PR description 

I felt compelled to take a stab at #84 and this is the result.

I believe it is functional as it is, though it could for sure use more testing. Wanted to see what you think before proceeding.

Some comments:
* Both ```Composite``` and ```CompositeTangent``` are templated on the underlying lie group type, i.e. ```CompositeTangent<double, SE3, R2>``` (and not ```CompositeTangent<double, SE3Tangent, R2Tangent>```) is the tangent of ```Composite<double, SE3, R2>```.
* Both types have a member function ```template<size_t I> get()``` that allows access to the individual parts via a ```Map```. I would have preferred to inherit from ```std::tuple``` to get ```std::get``` for free but it seems like it is assumed throughout the library that all types store the underlying data as a single continuous array.
* I first wrote it for C++17 and had to work quite hard to convert it to C++11. As a result there is a replacement for C++14 ```std::integer_sequence``` and the previous C++17 fold expressions are now using this [C++11 ```initializer_list``` trick](https://www.fluentcpp.com/2019/03/05/for_each_arg-applying-a-function-to-each-argument-of-a-function-in-cpp/).
* Some library parts are a bit unnatural for the composite type, such as the ```LieAlg``` and ```Transformation``` types. They are there so that the full interface is implemented and the tests run
* ```Jacobian``` and ```Transformation``` are block-diagonal and could potentially be represented with sparse instead of dense matrices.


artivis:
@pettni I opened this PR as a replacement to yours #175. I initially planned on opening a PR against your fork but there are conflicts due to this branch being rebased on devel. I thus went for the ~~lazier~~ simpler solution. 
If you're fine with that and this PR content, I'll merge it anytime soon. The main changes since last time I pinged you are listed in the PR description.

artivis:
While I didn't get any answer, I'm going to proceed merging this PR and closes #175.

PR #250, opened 2022-07-02
enable tests for autodiff real type

artivis:
These tests can finally be enabled since https://github.com/autodiff/autodiff/issues/185 was fixed.


artivis:
This PR requires [`autodiff`](https://github.com/autodiff/autodiff) >= 0.6.8 which currently does not build https://github.com/autodiff/autodiff/issues/233.

PR #271, opened 2023-06-02
replaced ::Constant(0) with ::Zero() due to ceres::Jet compile error

willat343:
I encountered the following compilation error when use the hat operator of the Rn group with ceres:
```
/usr/local/include/manif/impl/rn/RnTangent_base.h:125:34: error: no matching function for call to ‘Eigen::Matrix<ceres::Jet<double, 4>, 4, 4, 0, 4, 4>::Constant(int)’
  125 |   LieAlg t_hat = LieAlg::Constant(0);
```

I fixed this by replacing `::Constant(0)` with `::Zero()`. An alternative fix I think is to use `::Constant(Scalar(0))`. I chose the former because `::Zero()` is much more common in the codebase than `::Constant(Scalar(n))`.

I noticed 3 instances of `::Constant(Scalar(1))` in the code, in the SO2 base and tangent base classes. It would make sense to replace these with `::Ones()` but I have not included that in this pull request.

### DIFF
--- a/include/manif/impl/bundle/BundleTangent_base.h
+++ b/include/manif/impl/bundle/BundleTangent_base.h
@@ -366,7 +366,7 @@ struct GeneratorEvaluator<BundleTangentBase<Derived>>
   run(const unsigned int i, intseq<_Idx...>)
   {
     using LieAlg = typename BundleTangentBase<Derived>::LieAlg;
-    LieAlg Ei = LieAlg::Constant(0);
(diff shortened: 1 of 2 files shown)


artivis:
Thanks for those fixes :+1: .

artivis:
@willat343 could you please rebase your pr on top of `devel` ? Thanks.

willat343:
@artivis I think I have done so

artivis (approved):

PR #277, opened 2023-07-24
Fixes to the build: set the required C++ standard for tests; add a missing header

barracuda156:
This fixes two issues:

1. This: https://github.com/artivis/manif/issues/269 (actually fixes instead of discarding or leaving it to an end-user to fix).
2. Numerous errors due to a missing `<cassert>` header, like:

```
In file included from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_manif/manif/work/manif-bb3f6758ae467b7f24def71861798d131f157032/include/manif/impl/lie_group_base.h:6,
                 from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_manif/manif/work/manif-bb3f6758ae467b7f24def71861798d131f157032/include/manif/SO2.h:6,
                 from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_manif/manif/work/manif-bb3f6758ae467b7f24def71861798d131f157032/test/so2/gtest_so2.cpp:1:
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_manif/manif/work/manif-bb3f6758ae467b7f24def71861798d131f157032/include/manif/impl/so2/SO2_base.h: In instantiation of 'Eigen::Matrix<typename manif::LieGroupBase<Derived>::Scalar, 2, 1> manif::SO2Base<_Derived>::act(const Eigen::MatrixBase<OtherDerived>&, tl::optional<Eigen::Ref<Eigen::Matrix<typename manif::LieGroupBase<Derived>::Scalar, 2, 1> > >, tl::optional<Eigen::Ref<Eigen::Matrix<typename manif::LieGroupBase<Derived>::Scalar, 2, 2> > >) const [with _EigenDerived = Eigen::Matrix<double, 2, 1>; _Derived = manif::SO2<double>; typename manif::LieGroupBase<Derived>::Scalar = double; typename std::conditional<Eigen::Matrix<typename manif::LieGroupBase<Derived>::Scalar, 2, 1>::IsVectorAtCompileTime, Eigen::InnerStride<1>, Eigen::OuterStride<> >::type = Eigen::InnerStride<1>; typename std::conditional<Eigen::Matrix<typename manif::LieGroupBase<Derived>::Scalar, 2, 2>::IsVectorAtCompileTime, Eigen::InnerStride<1>, Eigen::OuterStride<> >::type = Eigen::OuterStride<>]':
[...]

PR #280, opened 2023-10-31
Set test target default cxx standard iif no glob set

artivis:
Set test targets default CXX standards iif no global standard is set by the user with e.g., `CMAKE_CXX_STANDARD` in the toolchain or `cmake -DCMAKE_CXX_STANDARD=xx` in the command line.

Replaces #277


artivis:
Hi @joansola, could you approve this PR so that I merge it? Thanks.

joansola (approved):

joansola:
good job :-)

PR #289, opened 2024-03-15
Add SGal(3)

artivis:
Add the Galilean Group SGal(3) as per,

  - "Absolute humanoid localization and mapping based on IMU Lie group and fiducial markers" Mederic Fourmy, Dinesh Atchuthan, Nicolas Mansard, Joan Sola, Thomas Flayols, [avilable at HAL](https://hal.science/hal-02183498/document)
  - "All About the Galilean Group SGal(3)" J. Kelly - [arxiv](https://arxiv.org/abs/2312.07555).

Todo:
- [x] left Jacobian
- [x] 'act' Jacobians


artivis:
@joansola FYI I've added SGal(3) to the [autodiff](https://github.com/autodiff/autodiff/) test suite.

joansola:
Regarding the left jac. Is the right jacobian ready, then? I think J.Kelly provides the left, and then we'll have to implement also the right via the trick `Jr(a)=Jl(-a)`?

artivis:
> Regarding the left jac. Is the right jacobian ready, then? I think J.Kelly provides the left, and then we'll have to implement also the right via the trick `Jr(a)=Jl(-a)`?

The rjac is implemented with respect to the left using the adj (see [here](https://github.com/artivis/manif/blob/00bc7211510d18bf8557ae85f869eaabde51f048/include/manif/impl/sgal3/SGal3Tangent_base.h#L182)).  
This can easily be replaced with `return (-*this).ljac();`.

You'll notice in the function some comments, I've implemented the 'easy blocks' of the right jac and validated those against auto diff.

joansola:
> This can easily be replaced with `return (-*this).ljac();`.

This would be a time saver over the Adj solution. And it works for all groups.

artivis:
> This would be a time saver over the Adj solution. And it works for all groups.

I'd have to check but I believe that all the other groups implement a closed-form anyway.

joansola:
> This can easily be replaced with `return (-*this).ljac();`.

We should definitely do this in sgal3

joansola:
This (in L247)
```c++
cB = (theta_sq + Scalar(2) * (Scalar(1) - theta * sin_t - cos_t)) / (Scalar(2) * theta_sq);
```

does not seem to reduce to this

```c++
cB = Scalar(1./8.) * theta_sq;
```

for small `theta`. I've checked several times. The problem is, every time I get a different result :-D. Will re-re-recheck

joansola:
Also, this statement has no `else` clause. What happens then if the `if` returns false? 

L262:
```c++
  if (theta_sq > Constants<Scalar>::eps) {
    cA = (Scalar(2) - theta * sin_t - Scalar(2) * cos_t) / theta_cu;
    cB = (theta_cu + Scalar(6) * theta + Scalar(6) * theta * cos_t - Scalar(12) * sin_t) / (Scalar(6) * theta_cu);
    cC = (Scalar(12) * sin_t - theta_cu - Scalar(3) * theta_sq * sin_t - Scalar(12) * theta * cos_t) / (Scalar(6) * theta_cu);
    cD = (Scalar(4) + theta_sq * (Scalar(1) + cos_t) - Scalar(4) * (theta * sin_t + cos_t) ) / (Scalar(2) * theta_cu);
    // cD = (Scalar(4) + theta_sq * (Scalar(1) + cos_t) - Scalar(4) * (theta * sin_t - cos_t) ) / (Scalar(2) * theta_cu);
    cE = (theta_sq + Scalar(2) * (cos_t - Scalar(1))) / (Scalar(2) * theta_cu);
    cF = (theta_cu + Scalar(6) * (sin_t - theta)) / (Scalar(6) * theta_cu);
  }

**Comment by joansola** [comment]
Implemented the else clause:

```c++
  if (theta_sq > Constants<Scalar>::eps) {
    cA = (Scalar(2) - theta * sin_t - Scalar(2) * cos_t) / theta_cu;
    cB = (theta_cu + Scalar(6) * theta + Scalar(6) * theta * cos_t - Scalar(12) * sin_t) / (Scalar(6) * theta_cu);
    cC = (Scalar(12) * sin_t - theta_cu - Scalar(3) * theta_sq * sin_t - Scalar(12) * theta * cos_t) / (Scalar(6) * theta_cu);
    cD = (Scalar(4) + theta_sq * (Scalar(1) + cos_t) - Scalar(4) * (theta * sin_t + cos_t) ) / (Scalar(2) * theta_cu);
    // cD = (Scalar(4) + theta_sq * (Scalar(1) + cos_t) - Scalar(4) * (theta * sin_t - cos_t) ) / (Scalar(2) * theta_cu);
    cE = (theta_sq + Scalar(2) * (cos_t - Scalar(1))) / (Scalar(2) * theta_cu);
    cF = (theta_cu + Scalar(6) * (sin_t - theta)) / (Scalar(6) * theta_cu);
  }else{
    cA = theta / Scalar(12);
    cB = theta_sq / Scalar(24);
    cC = theta_sq / Scalar(10);
    cD = theta_cu / Scalar(240);
    cE = theta / Scalar(24);
    cF = theta_sq / Scalar(120);
  }
```

Unfortunately, still no luck, and tests fail.

joansola:
please tell me if you want me to commit @artivis

joansola:
This is also weird: double assignation of matrix `S`:

```c++
  Ref33 S = Jl.template block<3, 3>(3, 3);        // first assign
  S = t() / Scalar(6) * V                         // second assign
    + (cA * W + cB * W * W) * (t() * V)
    // + cC * t() * (W * V * W)
    + cC * (W * V * (t() * W))
    // + cD * t() * (W * W * V * W)
    + cD * (W * W * V * (t() * W))
    + t() * V * (cE * W + cF * W * W)
    ;

  Jl.template block<3, 3>(0, 6) -= S;           // update
```

This seems wrong to me. Where is N1 in the (0,6) block of `Jl`?

In my opinion we should do:

```c++
Jl.template block<3,3>(0,6) = N1;     // <-- where is N1? Is it block (3,3)?
Ref 33 S = // bla bla all the formulas. S is N2 in Kelly's paper
Jl.template block<3,3>(0,6) -= S;    // update with N2

**Comment by joansola** [comment]
A little recap.

The tangent space is `[theta ; v ; rho ; t]`? please @artivis respond below:

- [ ] true
- [ ] false

Then, the error vector in the log() jacobian test is like so:

```
diff:
 9.8819340106803821e-05
-0.00037940868386976412
-5.7989384822931811e-05
-3.2089439994109625e-09
 -1.950976047382369e-09
-1.2978911279049044e-09
 3.0092006664261817e-10
  -1.27925553483621e-09
 4.7633497146648551e-10
                      0
```

that is, it is highest in the angular part (theta). Does that mean that the error lies in the angular matrix blocks of the Jacobian? Because then it is not in `N2`... how do you reason to indicate that the error is probably in N2?

**Comment by artivis** [comment]
Quick answers, I'll come back to it later.

- Good catch for `S`, it is indeed a mistake. Altho fixing it doesn't fix the tests.
- `N1` is computed and assigned [here](https://github.com/artivis/manif/blob/00bc7211510d18bf8557ae85f869eaabde51f048/include/manif/impl/sgal3/SGal3Tangent_base.h#L254)
- The tangent space is as in the paper, that is `[rho, nu, theta, iota]` (`[translational part, vel part, ang part, time part]`).
- If you look at the ceres-based test, `path/to/build/test/ceres/gtest_sgal3_ceres`, you'll see the whole Jacobian matrix compared against it's auto diff counter part. The diff is then computed element-wise on the matrices. That's my main indication that the blocks I pointed out earlier are the problem. Then given that `N1` is essentially `Q` in `SE3` and we already have this tested, I assume that for this block `N2` is to be scrutinized first.

**Comment by joansola** [comment]
OK it all makes sense. But I checked N1 many times already and still cannot find the bug.

**Comment by artivis** [comment]
> This trick of using 3x3 blocks of `Jl` as temporary placeholders is driving me crazy!

Jajaja, just replace them. We can optimize the code later ...

**Comment by joansola** [comment]
OK here's what I did:
- remove all temporary placeholders in `Jl`, replace with `Matrix<Scalar,3,3>`
- Name all 3x3 blocks D, E, L, M, N, N1 and N2 as in the paper
- Put computations in alfabetical order: D, E, L, M, N1, N2, N
- Add a bunch of comments

Still does not work. Will commit and continue from here.

**Comment by joansola** [comment]
Isn't it a bit confusing to have two versions of `asSO3()`, one in the Group and one in the Tangent? It is also the case for the other similar groups (SE3, SE_2_3)

**Comment by artivis** [comment]
Hi @joansola,
Maybe another lead for debugging: I added a couple more tests that currently fail for sgal3.  
Interestingly, one of them is [` delta.exp().inverse().adj() * delta.ljac() == (-delta).ljac() `](https://github.com/artivis/manif/blob/bbf1f16313c3fb801cdbaea4a788fd3e00418003/test/common_tester.h#L1173).

**Comment by joansola** [comment]
I suspect this is wrong:

```c++
template <typename _Derived>
void SGal3TangentBase<_Derived>::fillE(
  Eigen::Ref<Eigen::Matrix<Scalar, 3, 3>> E,
  const Eigen::Map<const SO3Tangent<Scalar>>& so3
) {
  using I = typename Eigen::DiagonalMatrix<Scalar, 3>;

  const Scalar theta_sq = so3.coeffs().squaredNorm();

  E.noalias() = I(Scalar(0.5), Scalar(0.5), Scalar(0.5)).toDenseMatrix();

  // small angle approx.
  if (theta_sq < Constants<Scalar>::eps) {
    return;
  }

  const Scalar theta = sqrt(theta_sq); // rotation angle

  const Scalar A = (theta - sin(theta)) / theta_sq;
  const Scalar B = (theta_sq + Scalar(2) * cos(theta) - Scalar(2)) / (Scalar(2) * theta_sq);

  const typename SO3Tangent<Scalar>::LieAlg W = so3.hat();

  E.noalias() += A * W + B * W * W;
}
```
since W is not normalized in the code, and it is in the paper (it is u^)

joansola:
> since W is not normalized in the code, and it is in the paper (it is u^)

This actually happens also in matrices `L`, `M` and `N2`

joansola:
OK BUGS solved!!! It was indeed the non-normalized `W` matrices in the code with respect to the normalized `u^` matrices in the paper.

Had to change the following:
- `fillE()` in `SGal3TangentBase.h`, used in `exp()` and block `E` of `ljac()`
- a verbose implementation of `fillE()` in `SGal3Base.h`, inside `log()`
- Computations of `L` and `N2` in `SGal3TangentBase.h`

Missing -- and thus not passing tests:
- `act()` 
- Jacobians of `act()`

joansola:
Regarding `act()`.

I see in Kelly's paper he speaks of the action on events, an event being a tuple [p,t] in R3 x R.

He also speaks of plenty of other possible actions: rotation, boost, and who knows what else.

Given that there are multiple possible actions, and we need to pick one, let us consider this:

- In SE(3), act is just transforming points with `T + R*p`
- In SE_2(3), act is doing the same as in SE(3).

So, in SGal(3) we could also keep doing the same as in SE(3). What do you think?

artivis:
@joansola all the tests are passing successfully :+1: .  
I now have to fix the Python bindings ...

joansola:
> @joansola all the tests are passing successfully 👍 . I now have to fix the Python bindings ...

Does this include the Jacobians of act()?

artivis:
I've got everything working fine :+1: .
I'll optimize a bit the Jac (e.g. use again those temporary blocks) and clean the history before doing a force-push.

joansola:
> e.g. use again those temporary blocks

yeah, you like freak coding :-) But please use `block <3,3>(6,0)` which is not used, otherwise it looks scary!

Also perhaps one can define all powers of theta up to `theta^5`, they are used multiple times. `theta^6` only once.

Do you think these optims matter so much?

include/manif/impl/sgal3/SGal3Tangent_base.h
```diff
@@ -0,0 +1,635 @@
...
+  // );
+
+  // Jr.template block<3, 3>(6, 6) = Jr.template topLeftCorner<3,3>();
+
+  // Jr(9, 9) = Scalar(1);
+
+  // Jr.template bottomLeftCorner<7, 3>().setZero();
+  // Jr.template block<4, 3>(6, 3).setZero();
+  // Jr.template block<1, 3>(9, 6).setZero();
+  // Jr.template block<6, 1>(3, 9).setZero();
```
joansola, line 205: Do we want all these coments?

include/manif/impl/sgal3/SGal3Tangent_base.h
```diff
@@ -0,0 +1,635 @@
...
+  /**
+   * @brief Get the small adjoint matrix ad() of SGal3
+   * that maps isomorphic tangent vectors of SGal3
+   * @return
+   */
+  Jacobian smallAdj() const;
+
+  // SGal3Tangent specific API
+
+  //! @brief Get the linear velocity part.
```
joansola, line 93: This is `rho`, the linear translation part

include/manif/impl/sgal3/SGal3Tangent_base.h
```diff
@@ -0,0 +1,635 @@
...
+  LinBlock lin();
+  const ConstLinBlock lin() const;
+
+  //! @brief Get the angular part.
+  AngBlock ang();
+  const ConstAngBlock ang() const;
+
+  //! @brief Get the linear acceleration part
```
joansola, line 101: this is `nu`, the linear velocity part.

Remember in IMU, we do refer to velocity and acceleration, but they are multiplied by Delta_t, becoming translation (aka position change) and velocity change.

include/manif/impl/sgal3/SGal3Tangent_base.h
```diff
@@ -0,0 +1,635 @@
...
+  //   Eigen::Ref<Eigen::Matrix<Scalar, 3, 3>> E,
+  //   const Eigen::Map<const SO3Tangent<Scalar>>& so3
+  // );
+
+  // static void fillS(
+    // Eigen::Ref<Eigen::Matrix<Scalar, 3, 3>> S,
+    // const Eigen::Map<const SO3Tangent<Scalar>>& so3,
+    // linearVel,
+    // t
+  // );
```
joansola, line 132: remove these commented bits. Also, in case we want to keep them, `S` was renamed `N1` in the implementation of `ljac()`

include/manif/impl/sgal3/SGal3Tangent_base.h
```diff
@@ -0,0 +1,635 @@
...
+  // block N1 = Q(rho,theta)
+  SE3Tangent<Scalar>::fillQ(Jl.template block<3, 3>(0, 6), rho_theta);
+
+  // Block N2, part of N
+  Scalar cC, cD, cE, cF;
+  if (theta_cu > Constants<Scalar>::eps) {
+    cA = (Scalar(2) - theta * sin_t - Scalar(2) * cos_t) / theta_cu / theta;
+    cB = (
+      theta_cu + Scalar(6) * theta + Scalar(6) * theta * cos_t - Scalar(12) * sin_t
+    ) / (Scalar(6) * theta_cu * theta_sq);
```
joansola, line 336: do we want to create pre-multiplied variables theta_four, theta_five? I leave it to you. So far we have theta_sq and theta_cu

joansola: theta_six is not necessary as it only appears once. But a `const` definition comes at no cpu cost and it's cleaner code.

include/manif/impl/sgal3/SGal3_base.h
```diff
@@ -0,0 +1,484 @@
...
+  template <typename _DerivedOther>
+  LieGroup compose(
+    const LieGroupBase<_DerivedOther>& m,
+    OptJacobianRef J_mc_ma = {},
+    OptJacobianRef J_mc_mb = {}
+  ) const;
+
+  /**
+   * @brief Get the action
+   * @param[in] v A 3D point.
```
joansola, line 95: could we call this point `p` and not `v`?

include/manif/impl/sgal3/SGal3_base.h
```diff
@@ -0,0 +1,484 @@
...
+   * @param[out] -optional- J_vout_m The Jacobian of the new object wrt this.
+   * @param[out] -optional- J_vout_v The Jacobian of the new object wrt input object.
+   */
+  template <typename _EigenDerived>
+  Eigen::Matrix<Scalar, 3, 1>
+  act(
```
joansola, line 101: I was thinking, regarding act() in all complex groups. Maybe we could have special actions. In case of SGal3, we would have:

act: T+Rp
boost: see paper
event-transform: see paper

and other possible actions only available for this group.

In SE3 we have a similar problem, in transforming points (T+R*p) and transforming vetors (R*v)

I'll add this comment as an issue in manif, for consideration.

joansola (changes requested):
Just consider the few comments in this review before merging. they are all easy.

include/manif/impl/sgal3/SGal3_base.h
joansola: Disregard this comment in this review.

include/manif/impl/sgal3/SGal3Tangent_base.h
artivis: Yeah I'll keep those as they are the actual closed-form of the right Jac. Now there are commented because of course it's missing some blocks.

include/manif/impl/sgal3/SGal3Tangent_base.h
artivis: My bad for copy/pasting.

include/manif/impl/sgal3/SGal3Tangent_base.h
artivis: Let's leave that to a potential sub-sequent pr

artivis:
Comments addressed :+1: .

include/manif/impl/sgal3/SGal3_base.h
```diff
@@ -92,14 +92,14 @@ struct SGal3Base : LieGroupBase<_Derived> {
 
   /**
    * @brief Get the action
-   * @param[in] v A 3D point.
+   * @param[in] p A 3D point.
    * @param[out] -optional- J_vout_m The Jacobian of the new object wrt this.
    * @param[out] -optional- J_vout_v The Jacobian of the new object wrt input object.
```
joansola, line 97: J_pout m
J_pout_p

joansola (changes requested):
still some minor change

include/manif/impl/sgal3/SGal3_base.h
artivis: Done.

joansola: J_pout_p
J_pout_m

see the `pout` instead of `vout`

joansola (approved):

artivis:
Thanks @joansola for your help on this PR!

joansola:
Visca!! Thanks to you @artivis

PR #300, opened 2024-07-24
Add Lie bracket

### LINKED ISSUE #267: Lie Bracket operation
I can't seem to find the Lie Bracket (sometimes called commutator) operation in any of the code, documentation or issues and wondered if this exists yet in manif, perhaps under a different name?

This is quite a hard blocker for my application (Manif with ceres, where i need the lie bracket operator in my computation).

If $A$ is a Lie algebra, then the lie bracket is is a mapping of form $[\cdot, \cdot]: A \times A \rightarrow A$. I noticed Sophus has a function `Tangent lieBracket(Tangent const& a, Tangent const& b)` which performs this operation with tangents instead.

As far as I know the lie bracket operation is unique to each lie algebra, so its not really something that is easy to implement in my code (I template for different lie groups) but would belong in manif. Most of the Lie brackets for common groups are not difficult to compute. Off the top of my head:
- For R^n lie algebra the lie bracket returns the zero vector
- For SO(3), the lie bracket returns (the hat of) the cross product of the tangent vectors
- For SE(3), I think the lie bracket is  (the hat of) the small adjoint of the first element multiplied with the second element as a tangent
- SO(2) and SE(2) should also have nice closed forms, but I don't have a textbook handy.

Is this something you would consider implementing, or perhaps already exists that I haven't found?

artivis:
Add the Lie bracket in vector form (` [x,y]^v = adj(x) . y `).

It is implemented as both : 
- ` x.bracket(y) `
- ` Tangent::Bracket(x, y)` 

with the later being a personal favorite in terms of notation.

The actual bracket in the Lie algebra can then be computed as `Tangent::Bracket(x, y).hat()`.

Note: the `vee` operator will be implemented in a subsequent pr.

Closes #267 .

### DIFF
--- a/include/manif/impl/bracket.h
+++ b/include/manif/impl/bracket.h
@@ -1,0 +1,35 @@
+#ifndef _MANIF_MANIF_IMPL_BRACKET_H_
+#define _MANIF_MANIF_IMPL_BRACKET_H_
+
+namespace manif {
+namespace internal {
+
+template <typename Derived>
+struct BracketEvaluatorImpl {
+  template <typename TL, typename TR>
+  static typename Derived::Tangent run(const TL& a, const TR& b) {
+    return a.smallAdj() * b;
+  }
+};
+
+template <typename Derived, typename DerivedOther>
+struct BracketEvaluator : BracketEvaluatorImpl<Derived> {
+  using Base = BracketEvaluatorImpl<Derived>;
+
+  BracketEvaluator(const Derived& xptr, const DerivedOther& xptr_o)
+  : xptr_(xptr), xptr_o_(xptr_o) {}
+
+  typename Derived::Tangent run() {
+    return Base::run(xptr_, xptr_o_);
+  }
+
+protected:
+
+  const Derived& xptr_;
+  const DerivedOther& xptr_o_;
+};
+
+} // namespace internal
+} // namespace manif
+
+#endif // _MANIF_MANIF_IMPL_BRACKET_H_
--- a/include/manif/impl/rn/RnTangent_base.h
+++ b/include/manif/impl/rn/RnTangent_base.h
@@ -202,6 +202,14 @@ struct RandomEvaluatorImpl<RnTangentBase<Derived>>
   }
 };
 
+template <typename Derived>
+struct BracketEvaluatorImpl<RnTangentBase<Derived>> {
+  template <typename TL, typename TR>
+  static typename Derived::Tangent run(const TL&, const TR&) {
+    return Derived::Tangent::Zero();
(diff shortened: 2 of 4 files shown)


artivis:
@joansola Your opinion about the api and more importantly about the fact that it is in vector form? Calling it 'bracket' isn't too much of a stretch since it is actually 'bracket^vee'?

Doing so has the advantage that it doesn't requires the `vee` operator (which will come in a subsequent pr).

nielsvd:
This is great, looking forward to seeing this merged. One question and one comment from my side:
* Are you planning to add specializations of `struct BracketEvaluatorImpl` in this PR already? Or is this planned future work?
* You may consider adding unit tests for the Lie bracket for specific Lie algebras, e.g., for SO3: `[V, W] == V.coeffs().cross(W.coeffs())`, for Rn and SO2: `[V, W] == 0`, and perhaps more.

joansola:
> @joansola Your opinion about the api and more importantly about the fact that it is in vector form? Calling it 'bracket' isn't too much of a stretch since it is actually 'bracket^vee'?
> 
> Doing so has the advantage that it doesn't requires the `vee` operator (which will come in a subsequent pr).

From a mathematical standpoint, yes, our vector representation of the Lie algebra is indeed a Lie algebra, and it has a bracket operator, which is this one, and they are both equivalent, because both spaces are isomorphic.

From a code standpoint. What´s the best way to document this so that a foreign can FIND this documentation trivially? Like, you are on vscode and you hover your mouse over. Then the doc appears and you can read this note just there. Because this is the only point of concern, it does not require a serious documentation, just a note. But this note needs to be findable very easily.

I´d say normally document it in the .h file as any other doxygen documentation. At least this. But being all templated, would the IDEs, by hovering over the function calls, display such doc? I dont know.

artivis:
@joansola Regarding documentation, hopefully the [doxygen+api](https://github.com/artivis/manif/pull/300/files#diff-466785fe7e5dc660b2d17e12833c62d901ae2814ed73880367dc093ac45266f8R259-R267) is clear enough and should do the job.

joansola:
```
L366   * @return @return LieAlg The Lie bracket [a,b] in vector form.
```

OK sounds good  but here you have two @return statements and say that **the return type is LieAlg** :-D

artivis:
@joansola Fixed the doc.

artivis:
@nielsvd Thanks for the feedback. 
I've implemented the specialisation for Rn :+1: . However I'll leave the group specific tests as futur work ;)

joansola (approved):
All seems good to me!!
:-)

PR #303, opened 2024-08-06
Remove extra semicolons after macros in Bundle.h

HenryLeC:
-Wpedantic logs wanings with extra semicolons


artivis (approved):
Thanks for the contributions :+1: .

PR #324, opened 2025-03-19
Revamp the documentation

### LINKED ISSUE #254: Documentation cover page looks bad in dark mode
In particular, the LaTeX symbols in the features table, which are black, are not visible if the browser is in dark mode.

### LINKED ISSUE #256: Links for c++ examples broken in documentation
In the documentation page:

https://artivis.github.io/manif/cpp/index.html

the links for the c++ examples are broken. For example, the link for the se2_localization.cpp file states:

https://artivis.github.io/manif/cpp/examples/se2_localization.cpp

while the true link should be e.g.:

https://github.com/artivis/manif/blob/devel/examples/se2_localization.cpp

maybe changing `devel` by some other branch.

### LINKED ISSUE #321: Document the order of the elements in the tangent space, in a prominent place
An important detail in the usage of manif is the meaning of the Jacobians. In particular, the order of the different elements in the tangent vector is crucial.

For example, in `SE(3)` we may have `[theta,p]` or `[p,theta]`. Mathematically, this is irrelevant. But code-wise, the order matters.

In SGal(3) there are 4 elements, so the number of options is huge, and only one is valid, It happens that the selected order (rho, nu, theta, t) does not match the order in the constructor of SGal3 (p, q, v, t), leading to possible confusion. The only good solution is a good documentation.

The arbitrary choice of these orderings should be clearly documented in a prominent place (e.g. in the GroupTangent classes), and appear in the documentation pages generated by doxygen.

I also noticed a few Error404 when browsing the doxygen documentation, especially when looking for specific files in the SGal3 class impl folders.

artivis:
Completely revamp the documentation.

Closes #254 
Closes #256 
Closes #321


artivis:
@joansola Could you please review/approve this PR ? But do *not* merge yet .

joansola:
Some things to final check before merge:

- [ ] The table of group operations is complete. For example, what about `vee`?
- [ ] The order of parts in the Jacobians is correct for all groups
    - [ ] SE2
    - [ ] SE3
    - [ ] SE23
    - [x] SGal3 
- [ ] Add an example of Ceres Jacobian compared to manif Jacobian, at least stating matrix sizes (not the actual math). For example, say that SO3d.Inverse() has a 3x3 Jac in manif, but since it is implemented by a quaternion, it has a 4x4 Jacobian if using `ceres::Jet`.
- [ ] I am not sure if the proper doc website should be complete by this PR. It seems it is lagging behind this PR by quite a bit, no?

joansola (commented):
see comment about last checks on the PR thread

artivis:
I'll double check the order of parts but I'm not going to add any new content to this PR. It was initially only for migrating from the old broken state to a new and shiny website and it's already be taking way more time than I had hoped. 
I still have to see if the actual deployment of this new website is working fine - that is, by merging this PR.
Feel free to open an issue where we can list all documentation misses, so that we address them in subsequent PRs.

joansola:
Sounds good to me! Sorry for being so picky :-D

joansola:
However, and sorry again @artivis . I checked the render in Safari and there's an issue. Maybe you want to merge anyway and I'll open an issue for this. But here it goes:

The operations table is not good. In the `devel` branch it is OK, but in the `fix/doc` branch here's how it looks like:

imagen

so most parts of the latex expressions are missing!

In Firefox it all looks good.

artivis:
> Sounds good to me! Sorry for being so picky :-D

Jajaja don't worry, on the contrary thanks for the effort :mechanical_arm: .

> I checked the render in Safari and there's an issue.

That's weird. I'll give it a quick search but it may have to wait for another PR.

joansola:
> That's weird. I'll give it a quick search but it may have to wait for another PR.

I agree to leave it for a separate PR.

artivis:
@joansola I've updated the `Tangent spaces` table. It is a little bit of a mess in restrospective :sweat_smile: .
I've also ordered the `group elements` column without specifying it since it's not exactly a one-to-one mapping (`R` != `q`).

Mind that I force-pushed to get rid of a conflict. You may want to pull clean to avoid any issue locally.

To me the PR is ready as is. If you agree (and approve this PR), I'll merge and we can eventually improve the doc in follow up PRs (e.g. there is plenty of work to make doxygen happy ...).

joansola (approved):
Revision OK!

artivis:
There it is, the new documentation website is live at [artivis.github.io/manif](https://artivis.github.io/manif/) !

joansola:
Great job @artivis !!!

PR #326, opened 2025-03-24
wip fix equations

artivis:


artivis:
Hi @joansola, I'm fixing the equations in the README using plain markdown - it now supports some latex-like math - see the branch [`fix/equations`](https://github.com/artivis/manif/tree/fix/equations). Should I also convert your 'tangent spaces' table ? Maybe only the two columns on the right ?

artivis:
> Should I also convert your 'tangent spaces' table ?

So, I went ahead and did just that. Have a look at this PR and let me know if that works for you.

joansola:
Yes it looks good! I think then that the ´group elements´column should also be passed to latex! I´ll do it right away.

joansola:
Done. @artivis see how it looks and for me this PR is ready.

joansola:
BTW @artivis did you check that the order in those tangent elements is OK with the actual order in manif? I think they are OK but I think you know better than me.

artivis:
I'll double check before the final merge :+1:

joansola:
@artivis the `\bm` directive does not render on my Firefox. Is is possible to add somewhere

```latex
\usepackage{bm}
```

?

artivis:
> @artivis the `\bm` directive does not render on my Firefox. Is is possible to add somewhere

What are you looking at? The README on github or the website locally? 
And what do you see? Simply the variable *not* bold or some error? 
It seems that github support for math is limited.

artivis:
Ok I've seen the `\bm` not rendering properly in the README on github. Looking into it.

artivis:
So, github markdown uses [mathJax](https://docs.mathjax.org/en/latest/index.html) for rendering math. 
MathJax supports the macros listed [here](https://docs.mathjax.org/en/latest/input/tex/macros/index.html).
`bm` isn't support but `boldsymbol` is. I've tried `boldsymbol`; locally it is fine, in github, it doesn't seem to work. The greek letters are rendered fine, just not bold ...

artivis:
I'll replace `bm` with `boldsymbol` since at least it doesn't produce an error. It also seems to work fine on the doc website.

artivis:
@joansola if this pr is fine with you, I'll go on and merge to the other pr which in turn will be merge as well to release the new doc website.

joansola:
The matrix Q is not rendering on Firefox, and the vector tau is a row vector.

image

joansola:
Same happens in:

Ubuntu Firefox
MacOS Firefox
MacOS Safari

so it's maybe a GitHub markdown issue.

It works fine on the vscode markdown renderer.

joansola:
The bold issue is not at all essential. I just wanted to idfferentiate scalars from vectors. But we can remove them all from the greek symbols if this works.

However, I tried to do so with the matrix Q, and it did not solve the not-rendering issue.

joansola:
OK I got it to work.

the underscore '\_' got scaped to mean "italic" and not subindex. I used "\\\_" instead of "\_" in the matrix Q, and now it works.

I remember this happened again to me sometime in Github using mathjax.

joansola:
I think this is not a neat solution, but it renders OK on firefox now.

It does not render well in vscode (now the underscore is visible, and the symbols are not sub-indexed), but htis is tolerable.

I doubt whether to validate this PR or not. What do you think? Any chance of having these underscores correctly interpreted by Github and/or mathjax?

joansola:
Oh, one thing to try (I have no time now):

the underscore issue happens when there are two underscores in one same line, since someone there interprets what's in between as an italics indicator.

Therefore, the solution is to breat the code of the matrix Q into a lot of different lines, each one containing only one underscore:

```
[...]
Q_xx &
Q_xy &
Q_xz \\
Q_yx &
[...]
```

etc

artivis:
The solution for this particular equation is to use the ` ```math ``` ` block instead of `$$ $$`.
I switched to double dollars because the `math` environment isn't recognised by the markdown engine of the doc website.
Let me fix this equation then I'll merge.

Edit: Your last suggestion doesn't work unfortunately.

artivis:
I finally came across another workaround, using a trailing space after the underscore [as suggested here](https://github.com/github/markup/issues/1575#issuecomment-1187437203). This renders fine on gh, vscode and the doc website.

artivis:
The matrix `Q` renders fine now, I'm merging this PR.

PR #331, opened 2025-06-04
Add all left and right jacobians

joansola:
Added a small PDF with a table with all left and right Jacobians


joansola:
@artivis can you review this PR? The table in the PDF has just been proofed by us at IRI.

PR #338, opened 2026-01-08
`SO2`/`SO3`: Rotation matrix constructors

ethanmusser:
Add constructors for the `SO2` and `SO3` classes that accept appropriately-sized rotation matrices and add corresponding Python bindings.

# Rationale

Without these constructors, users wishing to construct an  an `SO2`/`SO3` object from an Eigen rotation matrix must perform an intermediate conversion like the following.

```cpp
#include <Eigen/Geometry>
#include <manif/SO3.h>

int main()
{
  const Eigen::Matrix3d rot_mat{Eigen::Matrix3d::Identity()};
  const Eigen::AngleAxisd rot_vec{rot_mat};
  const manif::SO3d so3{rot_vec};
}
```

With this change, the following will work instead.

```cpp
#include <Eigen/Dense>
#include <manif/SO3.h>

int main()
{
  const Eigen::Matrix3d rot_mat{Eigen::Matrix3d::Identity()};
  const manif::SO3d so3{rot_mat};
}
```

# API Compatibility

I think that this is entirely new functionality and should not break any existing usages of the constructor since the new constructors only accept a rather restrictive concrete matrix type.  As such, anybody implicitly converting to an `Eigen::Quaternion` from an `Eigen::Matrix<Scalar, 4, 1>` or an `Eigen::AngleAxis` from an `Eigen::Matrix<Scalar, 3, 1>` should have that behavior preserved.

# Questions

- Is there any interest in having similar constructors for `SE2` and `SE3`?  Maybe something like the following.
  ```cpp
  SE2(const Scalar x, const Scalar y, const Eigen::Matrix<Scalar, 2, 2> &rotation_matrix);
  ```
  ```cpp
  SE3(const Translation &t, const Eigen::Matrix<Scalar, 3, 3> &rotation_matrix);
  ```
- Should the Python bindings include a check similar to [the one for the quaternion constructor](https://github.com/ethanmusser/manif/blob/4f31d8b97fcfba6bb9d21095e9ecfe71682c4ff2/python/bindings_so3.cpp#L30-L32) that ensures that the passed rotation matrix is both orthogonal and proper (i.e., `transpose(R) == inv(R)` and `det(R) == 1`)?

# Tags

- Most-recent editor: @artivis


ethanmusser:
Seems like the CI failures for `macos-14` and `windows-2022` might be related to the switch to Eigen3>=5.0.0.  Might have to do with what @traversaro is working on in #335.

ethanmusser:
@artivis Monthly ping to keep this from going stale.

